PR #27, opened 2025-05-27
feat: Add `ere-zisk` with only compile and execute utility.

han0110:
1. Implement `Compile` for `RV64_IMA_ZISK_ZKVM_ELF` by running command `cargo-zisk build`, and instead of returning compiled elf in bytes as other implementations, it only returns path to elf because `zisk` command (or sdk) only accept path to elf.
  Worth mentioning the command `cargo-zisk build` uses a global building directory which makes the `compile` panics if ran in parallel, we should fix this in futuer PR.
2. Implement `zkVM` for `EreZisk` by running commands of `cargo-zisk` directly.
  Worth mentioning, `zisk`'s guest program can only access the input as bytes, so the serilization format could differ from guest programs. I also change the `Input::as_bytes` to be fixed size encoding to make it deterministic, so the test program could be simpler.
3. Change the script to run the install script directly, and use `cargo-zisk` instead of `cargo zisk` because `cargo-zisk` didn't handle the first `zisk` argument.

As mentioned in 1. the compile couldn't be ran in parallel, to test `ere-zisk` we need to run `cargo test --package ere-zisk --lib -- --test-threads=1`

--- 

Some notes about my attepmt to use crates of [`zisk`](https://github.com/0xPolygonHermez/zisk) and [`proofman`](https://github.com/0xPolygonHermez/pil2-proofman) (instead of running command `cargo-zisk ...`):

1. No SDK like `sp1_sdk` to use directly, the logics are mostly scattered in [`cli`](https://github.com/0xPolygonHermez/zisk/tree/main/cli)'s sub-commands, so we need to copy the codes a lot.
2. The proving function panics in the end due to [this downcast operation](https://github.com/0xPolygonHermez/zisk/blob/main/cli/src/commands/prove.rs#L269-L273), the `result` is got from `witness_lib` which is installed along with `cargo-zisk`, but when I make the version same the downcast still panics, not sure how to fix this.


crates/ere-zisk/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[should_panic]
+    fn test_execute_zisk_no_input_for_guest_expecting_input() {
+        let elf_path = get_compiled_test_zisk_elf()
+            .expect("Failed to compile test ZisK guest for execution test");
+
+        let empty_input = Input::new();
+
+        let zkvm = EreZisk::new(elf_path);
+        let _ = zkvm.execute(&empty_input);
+    }
```
han0110, line 150: If the guest program panics or calls `std::process::exit` explicitly, `zisk`'s emulator will also panic with some memory reading issue, and the guest program's panic message is printed to stdout so we couldn't catch it easily, so here I use `#[should_panic]` for now but ideally we should check the execution error to be something like `HaltWithNonZeroExitCode`.

crates/ere-zisk/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+
+                if !status.success() {
+                    return Err(
+                        ZiskError::Prove(ProveError::CargoZiskProveFailed { status }).into(),
+                    );
+                }
+            }
+            ProverResourceType::Gpu => {
+                // TODO: Need to install another version of `cargo-zisk` with
+                //       `features = gpu` and call it here.
```
kevaundray, line 148: Can fix later

PR #29, opened 2025-05-30
zkvm-interface: add execution duration

jsign:
This PR adds a new field `execution_duration` to `ProgramExecutionReport`. This allows for later analytics around `cycles/second` or simply raw-zkVM raw execution performance.

### DIFF
--- a/crates/ere-openvm/src/lib.rs
+++ b/crates/ere-openvm/src/lib.rs
@@ -1,3 +1,5 @@
+use std::time::Instant;
+
 use openvm_build::GuestOptions;
 use openvm_circuit::arch::ContinuationVmProof;
 use openvm_sdk::{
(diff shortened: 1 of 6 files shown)


crates/zkvm-interface/src/reports.rs
```diff
@@ -10,13 +10,15 @@ pub struct ProgramExecutionReport {
     pub total_num_cycles: u64,
     /// Region-specific cycles, mapping region names (e.g., "setup", "compute") to their cycle counts.
     pub region_cycles: IndexMap<String, u64>,
+    /// Execution duration.
+    pub execution_duration: Duration,
 }
 
 impl ProgramExecutionReport {
     pub fn new(total_num_cycles: u64) -> Self {
```
jsign, line 18: Not that sure this ctor makes sense now that prob all impls can support setting the `elapsed_duration`.

I can also add it as a ctor parameter and use it that way. LMK

crates/ere-openvm/src/lib.rs
```diff
@@ -57,10 +59,7 @@ impl EreOpenVM {
     }
 }
 impl zkVM for EreOpenVM {
-    fn execute(
-        &self,
-        inputs: &Input,
-    ) -> Result<zkvm_interface::ProgramExecutionReport, zkVMError> {
+    fn execute(&self, inputs: &Input) -> Result<zkvm_interface::ProgramExecutionReport, zkVMError> {
```
kevaundray, line 62: This just seems to be cargo fmt

crates/zkvm-interface/src/reports.rs
kevaundray: We can add elapsed_duration too, but not a blocker for this PR. Using something like the builder pattern in the future, is likely the better API I think

PR #45, opened 2025-06-27
feat: add serde for Risc0Program

SuccinctPaul:
Once the Risc0Program can be cached, we can load_or_compile the guest program like sp1 did.


kevaundray (approved):
LGTM -- cc @han0110

han0110 (commented):
Need a rebase for the `Cargo.lock` then we could merge.

PR #47, opened 2025-07-01
feat: Impl zkVM for Nexus zkvm 

SuccinctPaul:
This pr implement zkvm-interface for Nexus zkvm, which leverage stwo as the default prover.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,8 +17,6 @@ members = [
 ]
 resolver = "2"
 
-exclude = ["tests"]
-
 [workspace.package]
 version = "0.1.0"
 edition = "2024"
--- a/crates/ere-nexus/build.rs
+++ b/crates/ere-nexus/build.rs
@@ -1,5 +1,5 @@
 use build_utils::detect_and_generate_name_and_sdk_version;
 
 fn main() {
-    detect_and_generate_name_and_sdk_version("jolt", "nexus-sdk");
+    detect_and_generate_name_and_sdk_version("nexus", "nexus-sdk");
 }
--- a/tests/nexus/guest/Cargo.toml
+++ b/tests/nexus/guest/Cargo.toml
@@ -10,4 +10,5 @@ nexus-rt = { git = "https://github.com/nexus-xyz/nexus-zkvm.git", tag = "0.3.4",
 #
 [features]
 cycles = [] # Enable cycle counting for run command
+[workspace]
 


kevaundray:
cc @han0110 for visibility

crates/ere-nexus/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    fn prove(
+        &self,
+        inputs: &Input,
+    ) -> Result<(Vec<u8>, zkvm_interface::ProgramProvingReport), zkVMError> {
+        let prover: Stwo<Local> = Stwo::new_from_file(&self.program.to_string_lossy().to_string())
+            .map_err(|e| NexusError::Prove(ProveError::Client(e.into())))
+            .map_err(zkVMError::from)?;
+
+        // let private_input =  vec![] ;
```
kevaundray, line 80: can you leave a comment for this?

One convention that may be useful for simplifying the design is that all inputs to the vm are private and all outputs are public. If an input should be public, then it could just be returned from the function.

crates/ere-zisk/src/lib.rs
```diff
@@ -187,7 +187,9 @@ impl zkVM for EreZisk {
                 unimplemented!()
             }
             ProverResourceType::Network(_) => {
-                panic!("Network proving not yet implemented for ZisK. Use CPU or GPU resource type.");
+                panic!(
+                    "Network proving not yet implemented for ZisK. Use CPU or GPU resource type."
+                );
```
kevaundray, line 192: This looks like an auto cargo fmt, so is fine

scripts/sdk_installers/install_nexus_sdk.sh
```diff
@@ -0,0 +1,54 @@
...
+# Verify Nexus installation
+echo "Verifying Nexus installation..."
+
+echo "Checking for RISC-V target..."
+if rustup target list | grep -q "riscv32i-unknown-none-elf"; then
+    echo "RISC-V target 'riscv32i-unknown-none-elf' not found."
+else
+    echo "RISC-V 'riscv32i-unknown-none-elf' not found after installation!" >&2
+    echo "Install the RISC-V target:"
+    rustup target add riscv32i-unknown-none-elf
```
kevaundray, line 44: Nice -- it uses one of the barebones targets. fyi, this might make it harder to use in our benchmarks for now since reth uses Arc in quite a few places for example, implying the atomic extension (unless it can be automatically lowered into normal loads and stores)

tests/nexus/guest/src/main.rs
```diff
@@ -0,0 +1,8 @@
+#![cfg_attr(target_arch = "riscv32", no_std, no_main)]
+
+use nexus_rt::print;
+
+#[nexus_rt::main]
+fn main() {
+    print!("Hello, World!\n");
```
kevaundray, line 7: Would be great to have a program that has inputs and it commits to some data as public input

crates/ere-nexus/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+        let mut private_input = vec![];
+        for input in inputs.iter() {
+            private_input.extend(
+                input
+                    .as_bytes()
+                    .map_err(|e| NexusError::Prove(ProveError::Client(e.into())))
+                    .map_err(zkVMError::from)?,
+            );
+        }
+
```
han0110, line 66: I think we missing `prover.run_with_input` here

tests/nexus/guest/src/main.rs
```diff
@@ -0,0 +1,18 @@
+#![cfg_attr(target_arch = "riscv32", no_std, no_main)]
+
+use nexus_rt::println;
+
+#[nexus_rt::main]
+#[nexus_rt::private_input(x)]
+fn main(x: u32) {
```
han0110, line 7: I think the input has to be `Vec<u8>` because we already serialize it outside of Nexus SDK, and we need to handle the deserialization from the result of `InputItem::as_bytes`.

crates/ere-nexus/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+        Ok(ProgramExecutionReport {
+            execution_duration: start.elapsed(),
+            ..Default::default()
+        })
```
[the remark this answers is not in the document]
kevaundray: > Doesn't catch execute in nexus

Could you maybe clarify and leave a comment?

crates/ere-nexus/src/lib.rs
han0110: Could you elaborate more what's that mean? You mean it doesn't count the execution cycle? If so we could leave it as is.

PR #48, opened 2025-07-08
Add automatic name and sdk version

jsign:
This PR adds a new mechanism that allows defining for each zkVM ere implementation:
- The name as a constant.
- The SDK version is automatically detected in the build phase.

This feature is used by `zkevm-benchmark-workload` to create a proper full name for the output results. For more information, see https://github.com/eth-act/zkevm-benchmark-workload/pull/104.

This PR is in draft mode to get early feedback. If it makes sense, it would be applied to the rest of the current zkVM implementations.

### DIFF
--- a/crates/build-utils/src/lib.rs
+++ b/crates/build-utils/src/lib.rs
@@ -2,19 +2,18 @@ use std::{env, fs, path::Path};
 
 // Generate a Rust source file that contains the name and version of the SDK.
 pub fn gen_name_and_sdk_version(name: &str, sdk_dep_name: &str) {
-    let workspace_dir = env::var("CARGO_WORKSPACE_DIR").unwrap();
-    let lock = fs::read_to_string(Path::new(&workspace_dir).join("Cargo.lock"))
-        .expect("Cargo.lock not found");
+    let meta = cargo_metadata::MetadataCommand::new()
+        .exec()
+        .expect("Failed to get cargo metadata");
 
-    let version = lock
-        .split("[[package]]")
-        .find(|pkg| pkg.contains(&format!("name = \"{sdk_dep_name}\"")))
-        .and_then(|pkg| {
-            pkg.lines()
-                .find(|l| l.trim_start().starts_with("version ="))
-                .and_then(|l| l.split('"').nth(1))
-        })
-        .unwrap_or_else(|| panic!("{sdk_dep_name} not found in Cargo.lock"));
+    let version = meta
+        .packages
+        .iter()
+        .find(|pkg| pkg.name.eq_ignore_ascii_case(sdk_dep_name))
+        .map(|pkg| pkg.version.to_string())
+        .unwrap_or_else(|| {
+            panic!("Dependency {sdk_dep_name} not found in Cargo.toml");
+        });
 
     let out_dir = env::var("OUT_DIR").unwrap();
     let dest = Path::new(&out_dir).join("sdk_version.rs");
--- a/crates/build-utils/src/lib.rs
+++ b/crates/build-utils/src/lib.rs
@@ -21,10 +21,10 @@ pub fn detect_and_generate_name_and_sdk_version(name: &str, sdk_dep_name: &str) 
 // Generate a Rust source file that contains the provided name and version of the SDK.
 pub fn gen_name_and_sdk_version(name: &str, version: &str) {
     let out_dir = env::var("OUT_DIR").unwrap();
-    let dest = Path::new(&out_dir).join("sdk_version.rs");
+    let dest = Path::new(&out_dir).join("name_and_sdk_version.rs");
     fs::write(
         &dest,
-        format!("pub const NAME: &str = \"{name}\";\npub const SDK_VERSION: &str = \"{version}\";"),
+        format!("const NAME: &str = \"{name}\";\nconst SDK_VERSION: &str = \"{version}\";"),
     )
     .unwrap();
     println!("cargo:rerun-if-changed=Cargo.lock");
--- a/crates/build-utils/src/lib.rs
+++ b/crates/build-utils/src/lib.rs
@@ -1,7 +1,7 @@
 use std::{env, fs, path::Path};
 
-// Generate a Rust source file that contains the name and version of the SDK.
-pub fn gen_name_and_sdk_version(name: &str, sdk_dep_name: &str) {
+// Detect and generate a Rust source file that contains the name and version of the SDK.
+pub fn detect_and_generate_name_and_sdk_version(name: &str, sdk_dep_name: &str) {
     let meta = cargo_metadata::MetadataCommand::new()
         .exec()
         .expect("Failed to get cargo metadata");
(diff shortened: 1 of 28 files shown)


crates/build-utils/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+
+// Generate a Rust source file that contains the name and version of the SDK.
+pub fn gen_name_and_sdk_version(name: &str, sdk_dep_name: &str) {
+    let workspace_dir = env::var("CARGO_WORKSPACE_DIR").unwrap();
+    let lock = fs::read_to_string(Path::new(&workspace_dir).join("Cargo.lock"))
+        .expect("Cargo.lock not found");
+
+    let version = lock
+        .split("[[package]]")
+        .find(|pkg| pkg.contains(&format!("name = \"{sdk_dep_name}\"")))
```
kevaundray, line 11: This LGTM, I haven't tried it, but maybe using `cargo_metadata::MetadataCommand;` from the cargo_metadata crate might be more robust?

jsign: I'll take a look!

crates/ere-sp1/src/lib.rs
```diff
@@ -13,6 +13,8 @@ use zkvm_interface::{
     ProverResourceType, zkVM, zkVMError,
 };
 
+include!(concat!(env!("OUT_DIR"), "/sdk_version.rs"));
```
kevaundray, line 16: This looks okay, in the future we could make it such that each zkVM needs to implement this by adding methods to the interface

kevaundray (approved):

crates/ere-sp1/src/lib.rs
jsign: Ah, adding `name()` and `sdk_version()` to some trait was my first option, but I wasn't sure if an existing one would be fine.

Do you think adding those methods to `zkVM` might be ok?

jsign: @kevaundray, another option is the `Compile` trait, but feels like from the existing traits `zkVM` (which contains `execute`, `prove`, `verify`) might be the best.

crates/build-utils/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+    gen_name_and_sdk_version(name, &version);
+}
+
+// Generate a Rust source file that contains the provided name and version of the SDK.
+pub fn gen_name_and_sdk_version(name: &str, version: &str) {
+    let out_dir = env::var("OUT_DIR").unwrap();
+    let dest = Path::new(&out_dir).join("name_and_sdk_version.rs");
+    fs::write(
+        &dest,
+        format!("const NAME: &str = \"{name}\";\nconst SDK_VERSION: &str = \"{version}\";"),
```
jsign, line 27: Now that `name()` and `sdk_version()` are methods in the `zkVM` trait, I removed the previous `pub`.

crates/build-utils/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+            panic!("Dependency {sdk_dep_name} not found in Cargo.toml");
+        });
+
+    gen_name_and_sdk_version(name, &version);
+}
+
+// Generate a Rust source file that contains the provided name and version of the SDK.
+pub fn gen_name_and_sdk_version(name: &str, version: &str) {
+    let out_dir = env::var("OUT_DIR").unwrap();
+    let dest = Path::new(&out_dir).join("name_and_sdk_version.rs");
```
jsign, line 24: Changed the file name to also include "name" in it.

crates/zkvm-interface/src/lib.rs
```diff
@@ -81,4 +81,10 @@ pub trait zkVM {
     /// TODO: We can also just have this return the public inputs, but then the user needs
     /// TODO: ensure they check it for correct #[must_use]
     fn verify(&self, proof: &[u8]) -> Result<(), zkVMError>;
+
+    /// Returns the name of the zkVM
+    fn name() -> &'static str;
+
+    /// Returns the version of the zkVM SDK (e.g. 0.1.0)
```
jsign, line 88: Clarified with an example to show we don't add the `v` prefix.

jsign (commented):
@kevaundray, prefered doing the trait style to be sure everyone is implementing this. Maybe you want to do a quick look again.

jsign:
Risc0 had the incorrect sdk name. SP1 was broken for other reasons (SDK mismatch between latest downloaded tooling and SDK in guest program).

Reg Pico, I think the latest release might be broken?

crates/ere-pico/Cargo.toml
```diff
@@ -8,8 +8,11 @@ license.workspace = true
 [dependencies]
 zkvm-interface = { workspace = true }
 thiserror = "2"
-pico-sdk = { git = "https://github.com/brevis-network/pico", tag = "v1.1.3" }
+pico-sdk = { git = "https://github.com/brevis-network/pico", tag = "v1.1.4" }
```
jsign, line 11: Tried this to fix Pico but looks like is still broken.

crates/ere-pico/src/lib.rs
```diff
@@ -77,10 +78,10 @@ impl zkVM for ErePico {
         }
 
         let start = Instant::now();
-        let num_cycles = client.emulate(stdin);
+        let emulation_result = client.emulate(stdin);
 
         Ok(ProgramExecutionReport {
-            total_num_cycles: num_cycles,
+            total_num_cycles: emulation_result.0,
```
jsign, line 84: Needed after the v1.1.4 update.

tests/sp1/compile/basic/Cargo.toml
```diff
@@ -6,4 +6,4 @@ edition = "2021"
 [workspace]
```
jsign, line 6: These SP1 example I pinned to tagged releases, since they were making the Docker builds fail, since this was pulling the default branch and the Docker downloads the latest tool tag and they weren't compatible.

Now everything points to the same tag, and the Docker build got fixed.

jsign:
@kevaundray, this ended up being a rabbit hole :laughing: 
Please take a look at the current unresolved comments!

kevaundray:
> @kevaundray, prefered doing the trait style to be sure everyone is implementing this. Maybe you want to do a quick look again.

Yep this looks good to me

crates/ere-pico/Cargo.toml
kevaundray: We can not build pico and create an issue -- it's not really possible to test it right now due to the memory overhead of CPU proving

kevaundray:
Cc @han0110 for visibility on  zisk

crates/ere-zisk/build.rs
```diff
@@ -0,0 +1,23 @@
...
+    if !out.status.success() {
+        panic!("cargo-zisk exited with {}", out.status);
+    }
+
+    // Example stdout:  "cargo-zisk 0.8.1 (f9a3655 2025-05-19T16:57:59.442084351Z)"
+    let text = std::str::from_utf8(&out.stdout).expect("cargo-zisk --version output");
+    let version = text
+        .split_whitespace()
+        .nth(1)
+        .expect("unexpected --version format");
```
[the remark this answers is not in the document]
kevaundray: Yeah I think doing it manually and then adding an issue would be easier -- we can always track down the zisk version if need be by looking at ere

kevaundray:
New changes and suggestions look okay -- feel free to liberally create issues on points above

crates/ere-pico/Cargo.toml
jsign: https://github.com/eth-act/ere/issues/49

jsign: Disabled it, so now looks like this https://github.com/eth-act/ere/actions/runs/16168531338
The workflow file has a reference to #49 to explain why.

PR #50, opened 2025-07-10
Add multi-gpu proving for risc0

han0110:
- Add script to install `cuda` for easier development
- Add script to install `docker` for easier development
- Remove `cuda` feature so we don't need feature flag to enable GPU proving
  - The building time is unaffected because the cuda GPU proving is using docker (image will be built when init the `EreRisc0`)
  - The `metal` feature remains optional because it increase the building time
- When the prover resource is GPU
  - `EreRisc0` builds the images if it doesn't exist (tagging with Risc0's SDK version), then use docker compose to spin up the proving services
  - Then send the image and input to the service for proving, and by interval (200ms) check if the proof is ready or not
  - When `EreRisc0` is dropped, it shuts down the services (if not properly dropped, the services get dangling.
  - In the future it'd be better to use channel to signal it successfully shut down the services, for caller to do gracefully shutdown.

Issues haven't figured out
- The first proving after the service is spin up will take some extra minutes, not sure what it's doing, or if there is any way to avoid this (just sleeping and waiting seems not working)

### DIFF
--- a/crates/ere-risc0/compose.yml
+++ b/crates/ere-risc0/compose.yml
@@ -1,4 +1,7 @@
-name: bento
+# Copied and modified from https://github.com/risc0/risc0/blob/v2.3.1/compose.yml.
+
+# MODIFIED: Use ere-risc0 as project name
+name: ere-risc0
 services:
   redis:
     hostname: ${REDIS_HOST}
@@ -44,42 +47,47 @@ services:
       timeout: 5s
       retries: 5
 
-  grafana:
-    image: ${GRAFANA_IMG}
-    restart: unless-stopped
-    ports:
-     - '3000:3000'
-    environment:
-      - GF_SECURITY_ADMIN_USER=admin
-      - GF_SECURITY_ADMIN_PASSWORD=admin
-      - GF_LOG_LEVEL=WARN
-      - POSTGRES_HOST=${POSTGRES_HOST}
-      - POSTGRES_DB=${POSTGRES_DB}
-      - POSTGRES_PORT={POSTGRES_PORT}
-      - POSTGRES_USER=${POSTGRES_USER}
-      - POSTGRES_PASS=${POSTGRES_PASSWORD}
-    volumes:
-      - ./grafana:/etc/grafana/provisioning/
-      - grafana-data:/var/lib/grafana
-    depends_on:
-      - postgres
-      - redis
-      - minio
+  # MODIFIED: Commented-out to avoid unnecessary resource waste.
+  # grafana:
+  #   image: ${GRAFANA_IMG}
+  #   restart: unless-stopped
+  #   ports:
+  #     - '3000:3000'
+  #   environment:
+  #     - GF_SECURITY_ADMIN_USER=admin
+  #     - GF_SECURITY_ADMIN_PASSWORD=admin
+  #     - GF_LOG_LEVEL=WARN
+  #     - POSTGRES_HOST=${POSTGRES_HOST}
+  #     - POSTGRES_DB=${POSTGRES_DB}
+  #     - POSTGRES_PORT={POSTGRES_PORT}
+  #     - POSTGRES_USER=${POSTGRES_USER}
+  #     - POSTGRES_PASS=${POSTGRES_PASSWORD}
+  #   volumes:
+  #     - ./grafana:/etc/grafana/provisioning/
+  #     - grafana-data:/var/lib/grafana
+  #   depends_on:
+  #     - postgres
+  #     - redis
+  #     - minio
 
   exec_agent:
-    image: agent
+    # MODIFIED: Specify the pre-built image.
+    image: ere-risc0/agent:${RISC0_VERSION}
     runtime: nvidia
+    # MODIFIED: Added to avoid image pulling.
+    pull_policy: never
     restart: always
     depends_on:
       - postgres
       - redis
       - minio
-    build:
-      context: .
-      dockerfile: bento/dockerfiles/agent.dockerfile
-      args:
-        NVCC_APPEND_FLAGS: "--gpu-architecture=compute_86 --gpu-code=compute_86,sm_86 --generate-code arch=compute_86,code=sm_86"
-        CUDA_OPT_LEVEL: 1
+    # MODIFIED: Commented-out because the image should be pre-built.
+    # build:
+    #   context: .
+    #   dockerfile: bento/dockerfiles/agent.dockerfile
+    #   args:
+    #     NVCC_APPEND_FLAGS: "--gpu-architecture=compute_86 --gpu-code=compute_86,sm_86 --generate-code arch=compute_86,code=sm_86"
+    #     CUDA_OPT_LEVEL: 1
     mem_limit: 4G
     cpu_count: 4
     environment:
@@ -91,15 +99,18 @@ services:
       S3_ACCESS_KEY: ${MINIO_ROOT_USER}
       S3_SECRET_KEY: ${MINIO_ROOT_PASS}
       RUST_LOG: ${RUST_LOG}
+      # MODIFIED: Added to set log color
+      NO_COLOR: ${NO_COLOR}
       RUST_BACKTRACE: 1
       RISC0_KECCAK_PO2: ${RISC0_KECCAK_PO2}
-      # Enable / disable along with gpu_*_agent*
-      # JOIN_STREAM: 1
-      # COPROC_STREAM: 1
+    # Enable / disable along with gpu_*_agent*
+    # JOIN_STREAM: 1
+    # COPROC_STREAM: 1
     entrypoint: /app/agent -t exec --segment-po2 ${SEGMENT_SIZE}
 
   aux_agent:
-    image: agent
+    # MODIFIED: Specify the pre-built image.
+    image: ere-risc0/agent:${RISC0_VERSION}
     runtime: nvidia
     pull_policy: never
     restart: always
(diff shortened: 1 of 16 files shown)


scripts/docker_image_builders/build_risczero_docker_image.sh
```diff
@@ -0,0 +1,20 @@
...
+git clone https://github.com/risc0/risc0.git "$DIR"
+cd "$DIR"
+
+# Build the images
+echo "Building bento images..."
+docker compose -f ./compose.yml --env-file ./bento/dockerfiles/sample.env build
+
+# Tag the services
+docker tag "agent" "ere-risczero/agent:latest"
+docker tag "bento-rest_api" "ere-risczero/rest_api:latest"
```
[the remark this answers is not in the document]
han0110: Because we are building image from source, and it doens't specify the image name of service `rest_api`, so the image name will be `{project}-rest_api` hence `bento-rest_api` (see https://github.com/risc0/risc0/blob/main/compose.yml#L179-L186). And it's `agent` instead of `bento-agent` is because it specifies the image name `agent`.

The `compose.yml` added in this PR removes the `build` part, so we can skip building on the runtime.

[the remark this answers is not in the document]
jsign: Ah, I see. The `compose.yml` in L13 is from the risc0 repo, then you tag it here to match the naming in _our_ compose file. Makes sense!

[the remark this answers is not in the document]
han0110: Moved the script to Rust code so the e2e process becomes simpler.

crates/ere-risc0/compose.yml
```diff
@@ -0,0 +1,118 @@
...
+  exec_agent:
+    image: ere-risc0/agent:${RISC0_VERSION}
+    runtime: nvidia
+    restart: always
+    depends_on:
+      - postgres
+      - redis
+      - minio
+    mem_limit: 4G
+    cpu_count: 4
```
han0110, line 58: This is copied from risc0's repo, not sure if we should keep it here. I tried to remove it and it didn't affect the performance.

crates/ere-risc0/src/prove/bento.rs
```diff
@@ -0,0 +1,230 @@
...
+        .map_err(|err| zkVMError::Other(err.into()))?;
+
+    loop {
+        let res = session
+            .status(&client)
+            .map_err(|err| zkVMError::Other(err.into()))?;
+
+        match res.status.as_ref() {
+            "RUNNING" => {
+                std::thread::sleep(Duration::from_millis(200));
```
han0110, line 56: By interval checking if a proof is ready or not, currently setting 200ms to not making the proving time too inaccurate, but I think in the future it'd be better to observe the container log or use other better way to do this.

scripts/sdk_installers/install_risc0_sdk.sh
```diff
@@ -62,7 +62,17 @@ fi
...
+
+if [[ -n "$RISC0_CLI_VERSION" && -n "$RISC0_CPP_VERSION" && -n "$RISC0_R0VM_VERSION" && -n "$RISC0_RUST_VERSION" ]]; then
+    # If versions are specified, install each component by their version
+    rzup install cargo-risczero $RISC0_CLI_VERSION
+    rzup install cpp $RISC0_CPP_VERSION
+    rzup install r0vm $RISC0_R0VM_VERSION
+    rzup install rust $RISC0_RUST_VERSION
+else
+    # Otherwise just install all components with by their latest version
+    rzup install
```
han0110, line 74: This installs the 4 components above with their latest released version.

docker/risc0/Dockerfile
```diff
@@ -5,6 +5,12 @@ FROM ${BASE_IMAGE_TAG}
 # Copy and run the Risc0 SDK installer script
 COPY scripts/sdk_installers/install_risc0_sdk.sh /tmp/install_risc0_sdk.sh
 
+# The install_risc0_sdk.sh script will respect these ENV variables.
+ENV RISC0_CLI_VERSION="2.3.1" \
+    RISC0_CPP_VERSION="2024.1.5" \
+    RISC0_R0VM_VERSION="2.3.1" \
+    RISC0_RUST_VERSION="1.88.0"
+
```
han0110, line 13: These are currently latest release.

crates/ere-risc0/src/prove/bento.rs
```diff
@@ -0,0 +1,230 @@
...
+                "clone",
+                "--single-branch",
+                "--branch",
+                &format!("v{SDK_VERSION}"),
+                "https://github.com/risc0/risc0.git",
+            ])
+            .arg(tempdir.path()),
+    )?;
+
+    cmd_exec_checked(
```
jsign, line 131: Coffee talk: so if we're using `ere-risc0` natively and we prove multiple things then:
- Each time we'll pull the Risc0 git repo in a temp directory, and
- Build the Risc0 images

I think pulling the git repo with `--depth 0` should be okay. I just wonder if the `docker compose build` will leverage the docker cache. Mostly because the second, third, etc runs will always happen in a new folder (but the context should be the same).

Anyway, just a thought... not sure if you have seen this confirmed in runs. :)

crates/ere-risc0/src/prove/bento.rs
```diff
@@ -0,0 +1,230 @@
...
+    docker_image_tag("bento-rest_api", rest_api_tag)?;
+
+    Ok(())
+}
+
+const BENTO_COMPOSE_TEMPLATE: &str = include_str!("../../compose.yml");
+const BENTO_GPU_PROVER_AGENT_TEMPLATE: &str = include_str!("../../gpu_prover_agent.yml");
+
+fn bento_compose_file() -> String {
+    let cuda_visible_devices = env::var("CUDA_VISIBLE_DEVICES").unwrap_or_else(|_| "".to_string());
```
jsign, line 151: If `ere-risc0` is run natively, this means the client must set this env var accordingly?

crates/ere-risc0/src/prove/bento.rs
```diff
@@ -0,0 +1,230 @@
...
+        .flat_map(|device_id| device_id.parse::<usize>().ok())
+        .collect::<Vec<_>>();
+
+    let mut compose: serde_yaml::Mapping = serde_yaml::from_str(BENTO_COMPOSE_TEMPLATE).unwrap();
+    let gpu_prover_agent: serde_yaml::Mapping =
+        serde_yaml::from_str(BENTO_GPU_PROVER_AGENT_TEMPLATE).unwrap();
+
+    if cuda_visible_devices.is_empty() {
+        let mut gpu_prover_agent = gpu_prover_agent.clone();
+        gpu_prover_agent.remove("deploy").unwrap();
```
jsign, line 163: This means this single container prover has access to all GPUs if the `CUDA_VISIBLE_DEVICES` isn't set?

crates/ere-risc0/src/prove/bento.rs
```diff
@@ -0,0 +1,230 @@
...
+    } else {
+        for idx in cuda_visible_devices {
+            let mut gpu_prover_agent = gpu_prover_agent.clone();
+            gpu_prover_agent["deploy"]["resources"]["reservations"]["devices"][0]["device_ids"]
+                [0] = idx.to_string().into();
+            compose["services"].as_mapping_mut().unwrap().insert(
+                format!("gpu_prove_agent{idx}").into(),
+                gpu_prover_agent.into(),
+            );
+        }
```
jsign, line 177: Cool, okay. I was craving to see this somewhere when reading the yamls before to make things click. :smile:

crates/ere-risc0/src/prove/bento.rs
```diff
@@ -0,0 +1,230 @@
...
+
+/// Execute `docker compose ... {command}` with `bento_compose_file()`.
+fn docker_compose_bento<I, S>(command: I) -> Result<(), io::Error>
+where
+    I: Clone + IntoIterator<Item = S>,
+    S: AsRef<OsStr>,
+{
+    let rust_log = env::var("RUST_LOG").unwrap_or_else(|_| "info".to_string());
+    let segment_size = env::var("SEGMENT_SIZE").unwrap_or_else(|_| "21".to_string());
+    let risc0_keccak_po2 = env::var("RISC0_KECCAK_PO2").unwrap_or_else(|_| "17".to_string());
```
jsign, line 191: Are these defaults coming from the Risc0 repo?

crates/ere-risc0/src/prove/bento.rs
```diff
@@ -0,0 +1,230 @@
...
+        .stdin(Stdio::piped())
+        .spawn()?;
+
+    let mut stdin = child.stdin.take().unwrap();
+    stdin.write_all(bento_compose_file().as_bytes())?;
+    drop(stdin);
+
+    let output = child.wait_with_output()?;
+    if !output.status.success() {
+        return Err(io::Error::other(format!(
```
jsign, line 209: I think this also means the `Drop` will cleanup stuff, no?

crates/ere-risc0/src/prove/bento.rs
```diff
@@ -0,0 +1,230 @@
...
+}
+
+/// Execute `docker compose ... up --detach` with `bento_compose_file()`.
+pub fn docker_compose_bento_up() -> Result<(), io::Error> {
+    docker_compose_bento(["up", "--detach", "--wait"])
+}
+
+/// Execute `docker compose ... down --volumes` with `bento_compose_file()`.
+pub fn docker_compose_bento_down() -> Result<(), io::Error> {
+    docker_compose_bento(["down", "--volumes"])
```
jsign, line 229: Nice, glad to see the volumes also being nuked.

crates/ere-risc0/compose.yml
```diff
@@ -0,0 +1,118 @@
...
+      - MINIO_ROOT_USER=admin
+      - MINIO_ROOT_PASSWORD=password
+      - MINIO_DEFAULT_BUCKETS=workflow
+    healthcheck:
+      test: ["CMD", "mc", "ready", "local"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+
+  exec_agent:
```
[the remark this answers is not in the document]
han0110: To be honest I haven't fully figured all the details, but from my current understanding, the `exec_agent` does the first pass of the execution and determine how many shards, and then `prover_agent` will pick the shard and recursive job to work on. About the `aux_agent` I'm not sure.

I also tried to remove the `runtime: nvidia` and it still works, but this is copied from risc0's repo so I didn't remove it.

crates/ere-risc0/compose.yml
```diff
@@ -0,0 +1,118 @@
...
+      S3_SECRET_KEY: password
+      RUST_LOG: ${RUST_LOG}
+      RUST_BACKTRACE: 1
+      RISC0_KECCAK_PO2: ${RISC0_KECCAK_PO2}
+    entrypoint: /app/agent -t exec --segment-po2 ${SEGMENT_SIZE}
+
+  aux_agent:
+    image: ere-risc0/agent:${RISC0_VERSION}
+    runtime: nvidia
+    pull_policy: never
```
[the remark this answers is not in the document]
han0110: hmm yeah it makes sense for `exec_agent` to also have this.

crates/ere-risc0/compose.yml
```diff
@@ -0,0 +1,118 @@
...
+  postgres:
+    hostname: postgres
+    image: postgres:16.3-bullseye
+    restart: always
+    environment:
+      POSTGRES_DB: taskdb
+      POSTGRES_USER: worker
+      POSTGRES_PASSWORD: password
+    expose:
+      - '5432'
```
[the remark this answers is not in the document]
han0110: Make sense, will test and check if it works.

> Not sure it is worth documenting any relevant modifications to the original?

Yes! I should have done this, I think initially I didn't intend to do so much modification, initially I only subsitute/hardcode some env variable, moving the `prover_agent` out as a template, remove the `snark_agent` that's not used in our case, and etc...

So I think it'd be worth to document our modification, perhaps we only commit the diff, and let `build.rs` to populate the 2 yml files, and don't subsitute the env by hand but also include the default env file.

crates/ere-risc0/src/lib.rs
```diff
@@ -29,40 +29,39 @@ impl Compiler for RV32_IM_RISC0_ZKVM_ELF {
...
+                    prove::bento::docker_compose_bento_up()
+                        .map_err(|err| zkVMError::Other(Box::new(err)))?;
+                }
             }
             ProverResourceType::Network(_) => {
                 panic!(
                     "Network proving not yet implemented for RISC Zero. Use CPU or GPU resource type."
                 );
             }
+            _ => {}
```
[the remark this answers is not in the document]
han0110: Make sense.

crates/ere-risc0/src/lib.rs
```diff
@@ -134,6 +128,14 @@ impl zkVM for EreRisc0 {
     }
 }
 
+impl Drop for EreRisc0 {
+    fn drop(&mut self) {
+        if matches!(self.resource, ProverResourceType::Gpu) && !cfg!(feature = "metal") {
+            prove::bento::docker_compose_bento_down().unwrap_or_else(|err| eprintln!("{err}"))
```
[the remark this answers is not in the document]
han0110: Make sense.

crates/ere-risc0/src/prove/bento.rs
```diff
@@ -0,0 +1,230 @@
...
+            .arg(src)
+            .arg(dst),
+    )
+}
+
+pub fn build_bento_images() -> Result<(), io::Error> {
+    let agent_tag = format!("ere-risc0/agent:{SDK_VERSION}");
+    let rest_api_tag = format!("ere-risc0/rest_api:{SDK_VERSION}");
+
+    if docker_image_exists(&agent_tag)? && docker_image_exists(&rest_api_tag)? {
```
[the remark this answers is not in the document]
han0110: The `SDK_VERSION` is from `build.rs`, in this case it'd be build time constant for both `ere-cli` and `ere-dockerzied` I think.

crates/ere-risc0/src/prove/bento.rs
```diff
@@ -0,0 +1,230 @@
...
+        return Ok(());
+    }
+
+    let tempdir = tempdir()?;
+
+    cmd_exec_checked(
+        Command::new("git")
+            .args([
+                "clone",
+                "--single-branch",
```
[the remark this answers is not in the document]
han0110: Ah yes! Didn't think about this.

crates/ere-risc0/src/prove/bento.rs
han0110: I think the only case we will build multiple time is by running `ere-risc0` (natively) or `ere-dockerized` risc0 (dockerized) for the first time, and instantiating multiple `EreRisc0` parallelly.

The reason is in line 113 it'd check if the image exist or not and early return, so ideally it'd only be built once. In my dev experience last few day I only need to build it once, and it took 20 mins so I guess I would have noticed if it builds again 🥲

But if in production env the local docker image gets lost every time, perhaps it'd be painful as you said.. In this case I guess it'd be worth to push this image to some registry to reuse it.

crates/ere-risc0/src/prove/bento.rs
han0110: Yes, it'd be ideal to set this for ere to know how many prover agent container to spin up.

If ran by `ere-dockerized` we also need to set this (and it'd be inherited properly)

crates/ere-risc0/src/prove/bento.rs
han0110: In this case if `CUDA_VISIBLE_DEVICES` is not set we assume there is no GPU so we don't give GPU access.

But it seems to make sense to give all GPU by setting `count: all`, since the resource is already assigned to be GPU.

crates/ere-risc0/src/prove/bento.rs
han0110: Will add some comment in the prover template to make it more straightforward.

crates/ere-risc0/src/prove/bento.rs
han0110: Yes, coming from their default env https://github.com/risc0/risc0/blob/v2.3.1/bento/dockerfiles/sample.env. The rest I already subsitute into the yaml file.

crates/ere-risc0/src/prove/bento.rs
han0110: Yes, and if the drop is not call properly, the containers dangles.. 

Actually I'm thinking to not detach to avoid the dangling problem, will try a bit and report back.

han0110 (commented):
Thanks for all the suggestions! Will try some ideas after I get back to my GPU machine (will ask for review again then, feel free to ignore this PR for now).

crates/ere-risc0/compose.yml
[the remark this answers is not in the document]
han0110: After tried that I realized if we'd still need the port mapping to make it accessible, because only the `ere-risc0` client will use `host` network if it's ran by `ere-dockerized`, the spin up docker compose services still use default `bridge` network.

han0110:
@jsign This is now ready for review

- I tried to reset the `compose.yml` and the modified diff can be found in https://github.com/eth-act/ere/pull/50/commits/cf9b33ae26a88117636adbf0d5272630d9f13fc2#diff-881625055c53741192bb7978abcc1793a84fcd8ee7928982171f78447947cf4c.
- Applied some suggestions in https://github.com/eth-act/ere/pull/50/commits/5f24aa312c409c0658d17da94fad9384761e8f87
- Add comments in about the 200ms in https://github.com/eth-act/ere/pull/50/commits/dc19a7bf25958bb3e451adf06f852cde64f2554c

Also filed an issue about the first proving time issue https://github.com/eth-act/ere/issues/81

Issue #14, opened 2025-05-24
Make Ere::compile use docker


kevaundray:
We can make ere use Docker, so that users won't need to install the compiler for each zkVM

PR #54, opened 2025-07-15
sp1: use Docker for guest program compilation

jsign:
This PR aims to determine the optimal approach for https://github.com/eth-act/ere/issues/14.
I took SP1 as the first target, but if convincing, we can apply the same idea to other zkVMs.

The main goal of the feature is to simplify `ere` users' experience by eliminating the need to install compilers on their machines, many of which have complex installation processes or can conflict with each other. 

The most obvious _user_ today is `zkevm-benchmark-workload` which has to compile many programs under different zkVMs.

~Note: still in draft since I'm testing this in `zkevm-benchmark-workload`.~ See [comment below](https://github.com/eth-act/ere/pull/54#issuecomment-3073619298).

### DIFF
--- a/crates/ere-sp1/src/compile.rs
+++ b/crates/ere-sp1/src/compile.rs
@@ -27,7 +27,7 @@ pub fn compile(guest_program_full_path: &Path) -> Result<Vec<u8>, CompileError> 
 
     info!("Compiling program: {}", guest_program_path_str);
 
-    Command::new("docker")
+    let status = Command::new("docker")
         .args([
             "run",
             "--rm",
@@ -43,7 +43,11 @@ pub fn compile(guest_program_full_path: &Path) -> Result<Vec<u8>, CompileError> 
             "/output",
         ])
         .status()
-        .map_err(|e| CompileError::DockerImageBuildFailed(Box::new(e)))?;
+        .map_err(CompileError::DockerCommandFailed)?;
+
+    if !status.success() {
+        return Err(CompileError::DockerContainerRunFailed(status));
+    }
 
     // Read the compiled ELF program from the output directory
     let elf = std::fs::read(elf_output_dir.path().join("guest.elf"))
--- a/crates/ere-sp1/src/error.rs
+++ b/crates/ere-sp1/src/error.rs
@@ -29,6 +29,10 @@ pub enum SP1Error {
 pub enum CompileError {
     #[error("Failed to build Docker image: {0}")]
     DockerImageBuildFailed(#[source] Box<dyn std::error::Error + Send + Sync + 'static>),
+    #[error("Docker command failed to execute: {0}")]
+    DockerCommandFailed(#[source] std::io::Error),
+    #[error("Docker container run failed with status: {0}")]
+    DockerContainerRunFailed(std::process::ExitStatus),
     #[error("Invalid guest program path: {0}")]
     InvalidGuestPath(PathBuf),
     #[error("Failed to create temporary directory: {0}")]


.github/workflows/rust-checks.yml
```diff
@@ -0,0 +1,56 @@
+name: Rust Checks
```
jsign, line 1: I removed `.github/workflows/test-sp1-docker.yml` for this new file.
The idea is now that since running tests also uses Docker for compiling guest programs, then the image building process is already indirectly tested.

And actually, in our previous workflows we didn't run any tests. Nor did some basic lint checks, so we have other benefits here.

I think as we migrate more zkVMs to use docker, we can simply keep editing this file to consider more and more `ere-*` crates.

crates/build-utils/src/docker.rs
```diff
@@ -0,0 +1,78 @@
...
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .status()
+        .is_err()
+    {
+        return Err(Error::DockerIsNotAvailable);
+    }
+
+    let cargo_workspace_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("../..")
```
jsign, line 22: When `ere` is used as a dependency, `CARGO_MANIFEST_DIR` will point to the path where the crate was downloaded by `cargo`. Then `../..` put us in the workspace place.

I think this works since the crate is pulled as a git repo? I'm not entirely sure.
If at some point we end up publishing crates in crate.io, we can figure out. If this doesn't work, we can embed the contents of Dockerfiles at compile time with `include_str!` and write stuff in temp files for Docker or similar.

crates/build-utils/src/docker.rs
```diff
@@ -0,0 +1,78 @@
...
+            "-t",
+            "ere-base:latest",
+            "-f",
+            dockerfile_base_path
+                .to_str()
+                .ok_or_else(|| Error::InvalidDockerfilePath(dockerfile_base_path.clone()))?,
+            cargo_workspace_dir.to_str().unwrap(),
+        ])
+        .status()
+        .map_err(|e| Error::DockerBuildFailed(e.into()))?;
```
jsign, line 41: Nothing interesting here. I double checked and parallel runs are fine.

crates/build-utils/src/docker.rs
```diff
@@ -0,0 +1,78 @@
...
+            "-t",
+            tag,
+            "-f",
+            dockerfile_path
+                .to_str()
+                .ok_or_else(|| Error::InvalidDockerfilePath(dockerfile_path.clone()))?,
+            cargo_workspace_dir.to_str().unwrap(),
+        ])
+        .status()
+        .map_err(|e| Error::DockerBuildFailed(e.into()))?;
```
jsign, line 59: A note about this in the `docker/sp1/Dockerfile` change below.

crates/ere-sp1/src/compile.rs
```diff
@@ -1,102 +1,53 @@
...
 
+use build_utils::docker;
 use tempfile::TempDir;
-use toml::Value as TomlValue;
 use tracing::info;
 
 use crate::error::CompileError;
 
-/// Compile the guest crate and return raw ELF bytes.
-pub fn compile_sp1_program(program_crate_path: &Path) -> Result<Vec<u8>, CompileError> {
```
jsign, line 10: All this got simplified.

crates/ere-sp1/src/compile.rs
```diff
@@ -1,102 +1,53 @@
...
-        elf_name
-    );
-
-    let status = Command::new("cargo")
-        .current_dir(program_crate_path)
+pub fn compile(guest_program_full_path: &Path) -> Result<Vec<u8>, CompileError> {
+    // Build the SP1 docker image
+    let tag = "ere-build-sp1:latest";
+    docker::build_image("docker/sp1/Dockerfile", tag)
+        .map_err(|e| CompileError::DockerImageBuildFailed(Box::new(e)))?;
```
jsign, line 13: This is to build the base and corresponding zkVM containers. It is a call that we'll do in all zkVMs. That comes below is zkVM specific ways to compile the guest program.

docker/sp1/Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN chmod +x /tmp/install_sp1_sdk.sh
 # The install_sp1_sdk.sh script will respect these ENV variables.
 # TODO: we are hardcoding /root which may not work for other users
 ENV SP1UP_HOME="/root/.sp1up" \
+    SP1UP_VERSION="v5.0.8" \
```
jsign, line 21: I changed the `install_sp1_sdk.sh` below to support this extra env flag.

If we don't do this, the SP1 docker image cache will become stale quite fast. Since "nothing change" future runs will use cached images since there's no way to distinguish if there's a new SP1 SDK as part of the build process.

Also is important to have reproducible builds. If we always download the "latest release", even with all this dockerization we can suddenly broken compilations if a new SDK release broke something (as happened some days ago both to me and @kevaundray)

docker/sp1/Dockerfile
```diff
@@ -34,11 +35,6 @@ RUN cargo prove --version
 # Copy the entire ere project context
 # The WORKDIR is /app from the base image
 WORKDIR /app
-COPY . .
-
-# Run tests
-RUN echo "Running tests for ere-sp1 library..." && \
-    cargo test --release -p ere-sp1 --lib -- --color always
```
jsign, line 41: No need to do this anymore since running tests using Docker is part of a GH workflow.

scripts/sdk_installers/install_sp1_sdk.sh
```diff
@@ -29,12 +29,14 @@ curl -L https://sp1up.succinct.xyz | bash -s -- --yes
 # and for subsequent commands if this script is sourced.
 export PATH="${SP1UP_HOME}/bin:${SP1_HOME}/bin:$PATH"
 
+export SDK_VERSION="${SP1UP_VERSION:-latest}"
```
jsign, line 32: Since this script is somewhat general, if the env var isn't provided we download the latest release as before.

jsign:
@kevaundray, testing this in `zkevm-benchmark-workload` I realized something interesting that makes sense in hindsight. 

The guest programs that we have in that repo, have in their `Cargo.toml` files indirect references to the workspace via `.workspace = true`. This make the compilation fail since we don't mount the whole workspace in the docker container.


I see two options:
- Allow as part of `ere` to separate "mount directory" from "guest program" directory. So we could mount the whole workspace and solve the problem, since now the guest program will find workspace references. 
  - Pro: solves the problem without `zkevm-benchmark-workload` changes
  - Cons: feels a bit hacky feature
- Leave `ere` "as is today" and expect guest programs to not be part of workspaces, so their `Cargo.toml` doesn't have any `.workspac`-like references
    - Pro: feels very clean from `ere` side.
    - Cons: forces `zkevm-benchmark-workload` guest programs `Cargo.toml` files to not reference workspace stuff; which can make maintaining correct matching versions a bit harder.

jsign:
> I thought we always have to mount the whole workspace (the first option), but if the guest workspace doesn't have dependency among members, we should be able to remove all workspace reference of members, but still get the benefit of small `/target` dir, if we maintain most of the dependency to have same versions. So it seems the second option is more optimal but might require some more effort to maintain (maybe negligible).

@han0110, I also feel Option 2. sounds cleaner despite the extra burden of having to maintain matching references between `ere-hosts` and the rest of guest programs. But maybe we can do some extra engineering to verify those things are matching in some CI workflow too, as a safety check. 

Let's see if @kevaundray has any strong pusbhack on Option 2. If not, I can make a `zkevm-benchmark-workspace` PR with the required changes for Option 2. and we can see how looks like.

docker/sp1/Dockerfile
jsign: Sorry, I think I put the wrong name here. This is the SDK version that will be used by `sp1up`. I should probably rename it to `SP1_SDK_INSTALL_VERSION`.

Does that clarify the confusion? Or maybe your comment still applies?

crates/ere-sp1/src/compile.rs
kevaundray: This was mainly here because different zkVMs have different ways to compile the program -- for SP1 it was straightforward because they have a CLI that could be used that also allows us to specify an output directory, when I was integrating the other zkVMs, some did not allow this

crates/ere-sp1/src/compile.rs
jsign: I'll add it in this PR :+1:

docker/sp1/Dockerfile
jsign: I renamed it.

crates/ere-sp1/src/compile.rs
jsign: Done.

jsign:
The ZISK CI seems broken again for a [new reason](https://github.com/eth-act/ere/actions/runs/16297405209/job/46023776708#step:5:999). I'll ignore it in this PR.

crates/ere-sp1/src/compile.rs
kevaundray: @jsign we can see what it looks like with the others implemented, I think we may need to bring this back and put a CLI over it and then this is what gets dockerized

kevaundray:
> > I thought we always have to mount the whole workspace (the first option), but if the guest workspace doesn't have dependency among members, we should be able to remove all workspace reference of members, but still get the benefit of small `/target` dir, if we maintain most of the dependency to have same versions. So it seems the second option is more optimal but might require some more effort to maintain (maybe negligible).
> 
> @han0110, I also feel Option 2. sounds cleaner despite the extra burden of having to maintain matching references between `ere-hosts` and the rest of guest programs. But maybe (if we're too paranoid) we can do some extra engineering to verify those things are matching in some CI workflow too, as a safety check.
> 
> Let's see if @kevaundray has any strong pusbhack on Option 2. If not, I can make a `zkevm-benchmark-workspace` PR with the required changes for Option 2. and we can see how looks like.

I think this makes sense -- longer term, we can fix the issue re having a small DSL that autogenerates the zkvm specific guest programs. Since most guest programs differ slightly

crates/ere-sp1/src/compile.rs
jsign: @kevaundray, ah, I see what you mean. Other zkVMs having more complex guest-compilation flows. Good point.

Uhm, maybe instead of dockerizing these Rust-like programs, we can simply make a `compile_guest_sp1.sh` scripts running exactly the same thing (without all the Rust overhead), and we mount also those scripts and simply execute the script.

So in the `compile_guest_xxx.sh` script we can be free to do whatever magic is needed for compiling guest programs as if you were doing it in a shell.

jsign:
BTW, interesting [thread convo here](https://github.com/eth-act/ere/pull/54#discussion_r2207982324). Looks like we might need to do another iteration to Dockerize guest compilers per zkVMs, since many of them aren't as simple as `cargo prove` style.

Summarized, sounds like the whole picture would be:
- Create one CLI per zkVM which has the logic to build a guest program (basically, the core that exist today in `compile(...)`). The CLI can assume all the toolchain is already installed in the machine.
- The CLI should be included in the corresponding `Dockerfile` which apart from installing the toolchain, also compiles the corresponding CLI binary and bakes it into the Docker image.
- The `compile(...)` in `ere` now can use the Dockerfile just calling that binary that is inside the docker image.

There are a couple of moving parts there -- maybe we can do that workflow for zkVMs that have complex enough compilation flows. For SP1 we could have the current simpler approach, but I suspect it might worth for all zkVMs to do the same for consistency even if that feels an overkill for SP1. 

cc @kevaundray @han0110

han0110:
The approach makes sense to me for most zkvms, since the toolchain is only necessary during compilation phase. And for `zisk` which doesn't provide Rust SDK to do `prove` and `verify`, I guess we could dockerize other methods of `zkVM` trait and call them by docker command from `ere-zisk`

kevaundray:
@jsign  I think we can merge this in for SP1 and then have followup PRs for the other crates

docker/sp1/src/main.rs
```diff
@@ -0,0 +1,103 @@
+use std::{fs, path::PathBuf, process::Command};
```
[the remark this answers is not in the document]
kevaundray: Not blocking; instead of docker/sp1 -- compilers/sp1 might be better in the future

PR #58, opened 2025-07-21
Risc0 compile by docker

han0110:
By commits:
1. Rename all naming `risczero` to `risc0` for consistency
1. Move `DockerRunCommand` from `ere-sp1` to `build-utils` to share between crates
1. Remove `ere-sp1`'s duplicated lib section
1. Duplicate the docker compilation process of `ere-sp1` for `ere-risc0`, and use the same CI job to test `ere-risc0`.


.github/workflows/rust-checks.yml
```diff
@@ -51,7 +51,17 @@ jobs:
       
       - name: Cache dependencies
         uses: Swatinem/rust-cache@v2
-      
+
+      - name: Install nvcc
```
jsign, line 55: Would this make sense in the GH CI? Since we can't do GPU proving tests, no?

I was thinking some time ago if we ever have some idle and cheap GPU prover machine, we could install the GH runner there and have some special CI workflow for GPU proving quick checks.

But until that happens not sure if installing nvcc makes sense? (maybe related on why the CI is failing)

jsign (approved):

.github/workflows/rust-checks.yml
han0110: Because the CI does `cargo clippy --all-features` so some of the dependencies of `risc0` has the `build.rs` script to build some cuda code, so I install `nvcc` to make the clippy happy.

I was thinking to remove the `--all-features` but I'm not sure if it's a good idea or not, so I temporarily disabled the `cuda` feature of `ere-risc0` so bypass that, since in https://github.com/eth-act/ere/pull/50 we will upgrade it to use docker the do the multi-gpu proving.

I can also remove the `--all-features` if there is no objection, sinice I didn't much features flags among `ere-zkvm` crates.

han0110: Then I think it'd be fine to remove `--all-features` because only `risc0` has feature flag.

crates/ere-risc0/src/compile.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    if !status.success() {
+        return Err(CompileError::DockerContainerRunFailed(status));
+    }
+
+    // Read the compiled ELF program from the output directory
+    let elf = std::fs::read(elf_output_dir.path().join("guest.elf"))
+        .map_err(CompileError::ReadCompiledELFProgram)?;
+
+    let image_id = compute_image_id(&elf).map_err(CompileError::ComputeImaegIdFailed)?;
```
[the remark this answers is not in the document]
han0110: fixed in https://github.com/eth-act/ere/pull/58/commits/57c42a35e69ffedb29af4f84d54a5ef3349af266

jsign:
> @jsign Curious whether we need the `prove` to run without any sdk involved (purely rust)? I think `risc0`'s local prover has some building process that takes very long (more than 30 mins), but if we install their sdk it will save a lot of time.

@han0110, you mean using Docker for `prove` as we do for the compilation? Or not using docker and assuming some SDK is installed in the host machine?

If the latter, I would prefer to avoid that, as adding assumptions about SDKs being installed on the host machine to use `ere` would feel like a step backward (i.e., mainly why we dockerized the compilation step). Also fear a bit if we try using the same machine to prove multiple zkVMs and multiple have conflicting SDKs somehow.

han0110:
> Using docker for proving I think sounds fine -- I think it should be OK with how the multi-gpu mechanics of Risc0 works? (I guess would use Docker in Docker?)

yeah risc0's multi-gpu also uses docker, in that case I guess we don't need to dockerize it again, but for cpu prover it doesn't provide a dockerized version.

I'll add the dockerized prove soon, and ask for review when it's finished.

crates/ere-risczero/build_script_template.rs
```diff
@@ -1,62 +0,0 @@
-// This is ere-risczero/build_script_template.rs
```
kevaundray, line 1: Just trying to figure out why this is not needed anymore

han0110: Preivously we use the `build.rs` to call the [`embed_methods`](https://github.com/risc0/risc0/blob/main/risc0/build/src/lib.rs#L844-L846), but it seems easier to just call the `cargo risczero build` which also [outputs the elf path and image id](https://github.com/risc0/risc0/blob/main/risc0/cargo-risczero/src/commands/build.rs#L58), so we could remove this.

crates/ere-risc0/src/compile.rs
```diff
@@ -0,0 +1,111 @@
...
+    path::{Path, PathBuf},
+    str::FromStr,
+};
+use tempfile::TempDir;
+use tracing::info;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Risc0Program {
+    // TODO: Seems like the risc0 compilation is also compiling
+    // TODO: the analogous prover and verifying key
```
kevaundray, line 15: Can you make an issue for this?

kevaundray (approved):

crates/ere-risc0/src/compile.rs
han0110: Ah actually I didn't know what this mean, it's just moved from the original compile logic.

Could you elaboate more about this? Because the output of `cargo risczero build` seems to be only the elf, I didn't see the prover and verifying key.

han0110: I see, as far as I see I the build command just calls `cargo build ...` internally, doens't do other preprocess thing, so I think it'd be fine to remove these comments (will do so in futuer PR)

PR #62, opened 2025-07-22
Update SP1 SDK to v0.5.8

jsign:


crates/ere-sp1/Cargo.toml
```diff
@@ -8,7 +8,7 @@ license.workspace = true
 [dependencies]
 zkvm-interface = { workspace = true }
 build-utils.workspace = true
-sp1-sdk = "5.0.5"
+sp1-sdk = "5.0.8"
```
jsign, line 11: Note that the Dockerfile version was [correctly pinned](https://github.com/eth-act/ere/blob/master/docker/sp1/Dockerfile#L30) from some days ago.

PR #66, opened 2025-07-25
Add crate `ere-dockerized`

han0110:
- Add crate `ere-dockerized` which provides:
  - A `cli` to do `compile`, `execute`, `prove` and `verify` of other `zkVM` implementations
  - `EreDockerized` which implements `zkVM` by calling dockerized `cli`
- Clean up the usage of `docker` in `ere-sp1` and `ere-risc0`
- Features `sp1`, `risc0` and `zisk` works in `ere-dockerized`, but `ere-zisk` requires too much disk space so we don't test it in CI.

## Rationale

If the `ere-{zkvm}` user doesn't want to install the zkvm's SDK, but happy to use docker, the user can use `EreDockerizedCompiler(zkvm)` as the `Compiler` and `EreDockerizedzkVM::new(zkvm, program, resource)` as the `zkVM` directly.

If the user can guarantee to have installed the SDK, the user can use `ere-{zkvm}` to avoid the docker overhead.

This helps us to separate the native impl and docker impl in more straightforward way I believe.

## Todos

- [ ] Currently in CI we only test the `ere-dockerized`, should we also test other native crates natively (runs the SDK installer beforehand)
- [ ] In order to be supported in `cli`, we need the `Compiler::Program` to be de/serializable, currently `jolt` and `openvm`'s program doesn't satisify that, hence doesn't compile.
- [ ] We need to make sure at least `zkVM::verify` method could be compiled to wasm if underlying zkvm supports, so we could re-organize the crates into `ere-{zkvm}/native` and `ere-{zkvm}/generic`, the former is current implementation that can depend on pre-installed binaries, the latter should be able to be compiled to wasm and portable.


jsign (approved):
LGTM, great work! Left some comments for your consideration.

crates/ere-cli/src/main.rs
```diff
@@ -0,0 +1,259 @@
...
+        proof_path: PathBuf,
+        /// Path where the report will be written
+        report_path: PathBuf,
+    },
+    /// Verify execution of a compiled program
+    Verify {
+        /// Path to the compiled program
+        program_path: PathBuf,
+        /// Path to the proof
+        proof_path: PathBuf,
```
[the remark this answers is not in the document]
kevaundray: I think I've gone back and forth on this; the main issue being that the proof systems likely want them to be combined, but the user who verifies the proof will then need to separate them to verify the public inputs against something they computed.

One idea is to have the proof come with public inputs, and the verify method returns both whether the proof was valid and the public inputs assosciated with the proof. This means that its on proof system/Ere to know how to layout the proof with public input and how to separate them

crates/ere-cli/src/main.rs
```diff
@@ -0,0 +1,259 @@
...
+        Ok(ere_nexus::EreNexus::new(program, resource))
+    }
+
+    #[cfg(feature = "openvm")]
+    {
+        let program =
+            bincode::deserialize(&program).with_context(|| "Failed to deserialize program")?;
+        ere_openvm::EreOpenVM::new(program, resource)
+            .with_context(|| "Failed to instantiate EreOpenVM")
+    }
```
kevaundray, line 230: Could we create an issue, re this: If we add a bunch of zkVMs, then adding them all into this function doesn't really scale

kevaundray:
Can any of this be pulled into smaller PRs? Maybe ere-dockerized for example

crates/ere-dockerized/build.rs
```diff
@@ -0,0 +1,58 @@
...
+    .try_into()
+    .unwrap();
+    let zisk_version = "0.9.0";
+
+    let zkvm_sdk_version_impl = format!(
+        r#"impl crate::ErezkVM {{
+    pub fn sdk_version(&self) -> &'static str {{
+        match self {{
+            Self::Jolt => "{jolt_version}",
+            Self::Nexus => "{nexus_version}",
```
kevaundray, line 44: Noting: If a new zkVM is added, then this _should_ fail and we can update it accordingly

crates/ere-dockerized/src/docker.rs
```diff
@@ -0,0 +1,279 @@
...
+        let Self(key, value) = self;
+        match value {
+            Some(value) => write!(f, "--{key}={value}"),
+            None => write!(f, "--{key}"),
+        }
+    }
+}
+
+#[derive(Clone)]
+pub enum DockerInstruction {
```
kevaundray, line 34: Instead of building a dockerfile using DockerInstructions, would we have less of a diff, if we had a template Dockerfile and then just "filled in" the parts we wanted?

```rust
ENV {{placeholder}}
```

For example

kevaundray:
> > Can any of this be pulled into smaller PRs? Maybe ere-dockerized crate only for example
> 
> Makes sense, I also start to feel difficult to handle the feedbacks in this case, I'll try to split it into 3 parts (includes your feedbacks in this PR):
> 
> * Add `ere-cli`
> * Add `ere-dockerized`
> * CI refactors + some fixes
> 
> @jsign @kevaundray I'll ask for your reviews later, pardon me for creating this huge PR 🙏

That would be great ser and no worries! Big PRs are usually good for showing the end to end thing for those with high context, though for reviewing, smaller PRs are better -- we can probably use this PR when reviewing the smaller PRs to see how it all fits together

PR #67, opened 2025-07-28
Do preprocess when instantiating `EreOpenVM`

han0110:
Also change OpenVM's `Compiler::Program` to raw ELF bytes, so it's de/serializable (required by #66).

### DIFF
--- a/crates/ere-openvm/src/lib.rs
+++ b/crates/ere-openvm/src/lib.rs
@@ -1,18 +1,19 @@
-use std::{path::Path, time::Instant};
-
+use crate::error::{CompileError, ExecuteError, OpenVMError, VerifyError};
 use openvm_build::GuestOptions;
-use openvm_circuit::arch::ContinuationVmProof;
+use openvm_circuit::{
+    arch::{ContinuationVmProof, instructions::exe::VmExe},
+    system::program::trace::VmCommittedExe,
+};
 use openvm_sdk::{
-    Sdk, StdIn,
+    F, SC, Sdk, StdIn,
     codec::{Decode, Encode},
-    config::{AppConfig, SdkVmConfig},
-    prover::AppProver,
+    config::{AppConfig, DEFAULT_APP_LOG_BLOWUP, DEFAULT_LEAF_LOG_BLOWUP, SdkVmConfig},
+    keygen::AppProvingKey,
 };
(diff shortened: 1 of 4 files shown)


crates/ere-openvm/src/lib.rs
```diff
@@ -1,78 +1,133 @@
...
+            let toml = fs::read_to_string(&app_config_path).map_err(|source| {
+                CompileError::ReadConfigFailed {
+                    source,
+                    path: app_config_path.to_path_buf(),
+                }
+            })?;
+            toml::from_str(&toml)
+                .map_err(|err| CompileError::DeserializeConfigFailed(err.into()))?
+        } else {
+            AppConfig {
```
jsign, line 68: Do you have any thoughts on why defining this default here and make it part of the program serialization, compared to make it a static var used in `new`? No strong opinions, just curious.

jsign (approved):
LGTM!

crates/ere-openvm/src/lib.rs
han0110: I feel the app config is part of compilation (compilation flags thing), and I guess for performance reason most project will have one (to enable certain precompiles), so I guess it'd be okay to store it in `Program`.

If it's a argument in `new`, I'm afraid it'd make host a bit annoying because it needs to check if the guest dir has this config or not.

crates/ere-openvm/src/lib.rs
```diff
@@ -1,78 +1,133 @@
...
+                app_fri_params: FriParameters::standard_with_100_bits_conjectured_security(
+                    DEFAULT_APP_LOG_BLOWUP,
+                )
+                .into(),
+                app_vm_config: SdkVmConfig::builder()
+                    .system(Default::default())
+                    .rv32i(Default::default())
+                    .rv32m(Default::default())
+                    .io(Default::default())
+                    .build(),
```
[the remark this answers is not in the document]
han0110: Yes, by default it's RISCV32IM with IO but no precompiles, and the default app config is copied from https://github.com/openvm-org/openvm/blob/main/crates/cli/src/default.rs#L31.

Added a comment above it to clarify in this commit https://github.com/eth-act/ere/pull/67/commits/cd2402e38e16ccc99fb07c4c5a655195fab5bb76, thanks for pointing out.

PR #68, opened 2025-07-28
Do preprocess when instantiating `EreJolt`

han0110:
- Change Jolt's `Program` to `PathBuf` because `jolt::host::Program` only takes path to elf.
- Add `EreJoltProof` for easier de/serialize
- Replace unwraps with errors


crates/ere-jolt/src/lib.rs
```diff
@@ -24,53 +28,73 @@ pub struct JOLT_TARGET;
 impl Compiler for JOLT_TARGET {
     type Error = JoltError;
 
-    type Program = Program;
+    // Note that we use `PathBuf` as `Program` because the only way to pass elf
+    // information to `jolt::host::Program` is by setting elf path, so we don't
+    // bother to read it into memory here.
+    type Program = PathBuf;
```
jsign, line 34: I guess should be fine if the compilation and further use of the program happen close in time (imagining if something could delete that file), and always on the same machine.

crates/ere-jolt/src/lib.rs
```diff
@@ -24,53 +28,73 @@ pub struct JOLT_TARGET;
...
 }
 
 impl EreJolt {
-    pub fn new(
-        program: <JOLT_TARGET as Compiler>::Program,
-        _resource_type: ProverResourceType,
-    ) -> Self {
-        EreJolt { program }
+    pub fn new(elf_path: PathBuf, _resource: ProverResourceType) -> Self {
+        let mut program = Program::new("");
```
jsign, line 78: When compiling in L52 it used the package name. Is it ok to use the empty string here?

crates/ere-jolt/src/lib.rs
```diff
@@ -24,53 +28,73 @@ pub struct JOLT_TARGET;
...
     ) -> Result<zkvm_interface::ProgramExecutionReport, zkVMError> {
-        // TODO: check ProgramSummary
-        // TODO: FIXME
-        // let summary = self
-        //     .program
-        //     .clone()
-        //     .trace_analyze::<jolt::F>(inputs.bytes());
-        // let trace_len = summary.trace_len();
-        let trace_len = 0;
+        // TODO: Check how to pass private input to jolt
```
jsign, line 95: Side question: Do we want to pass private inputs to zkVMs in ere? Prob we need to work a bit the `Input` type to reflect this?

crates/ere-openvm/src/lib.rs
```diff
@@ -1,78 +1,133 @@
-use std::{path::Path, time::Instant};
```
jsign, line 1: From here below is just because you branched from https://github.com/eth-act/ere/pull/67 ?

jsign (approved):

jsign:
BTW, I haven't interacted with the Jolt SDK in detail, so maybe you'd like a further opinion from @kevaundray. Take my review as somewhat light here. :)

crates/ere-openvm/src/lib.rs
han0110: Yes, it was a mistake, just rebased it so it doesn't depend on it.

crates/ere-jolt/src/lib.rs
jsign: Sorry, I was thinking about zk. Yes, all inputs must be private.

PR #71, opened 2025-07-31
Add `ere-cli`

han0110:
First part of #66 

Implement `ere-cli` which provides a cli to do `compile`, `execute`, `prove` and `verify` of specific zkVM implementation selected by feature flag. The reason to not include all `zkvm` is due to compliation time, in case one only wants to use single zkvm, not sure if this is the best approach so open to disuccsion.

The imagined usage of the cli is with `ere-dockerized`, which handles the de/serialization of program, input, etc... It seems not very easy to run it by hand (due to serialization)

### DIFF
--- a/crates/ere-cli/src/main.rs
+++ b/crates/ere-cli/src/main.rs
@@ -1,0 +1,213 @@
+use anyhow::{Context, Error};
+use clap::{Parser, Subcommand};
+use std::{fs, path::PathBuf};
+use zkvm_interface::{Compiler, ProverResourceType, zkVM};
+
+mod serde;
+
+// Compile-time check to ensure exactly one backend feature is enabled
+const _: () = {
+    assert!(
+        (cfg!(feature = "jolt") as u8
+            + cfg!(feature = "nexus") as u8
+            + cfg!(feature = "openvm") as u8
+            + cfg!(feature = "pico") as u8
+            + cfg!(feature = "risc0") as u8
+            + cfg!(feature = "sp1") as u8
+            + cfg!(feature = "zisk") as u8)
+            == 1,
+        "Exactly one zkVM backend feature must be enabled"
+    );
+};
+
+#[derive(Parser)]
+#[command(author, version)]
+struct Cli {
+    #[command(subcommand)]
+    command: Commands,
+}
+
+#[derive(Subcommand)]
+enum Commands {
+    /// Compile a guest program
+    Compile {
+        /// Path to the base directory (workspace root)
+        #[arg(long)]
+        mount_directory: PathBuf,
+        /// Relative path from `mount_directory` to the guest program
+        #[arg(long)]
+        guest_relative: PathBuf,
+        /// Path where the compiled program will be written
+        #[arg(long)]
+        program_path: PathBuf,
+    },
+    /// Execute a compiled program
+    Execute {
+        /// Path to the compiled program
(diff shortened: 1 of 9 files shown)


crates/ere-cli/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+            + cfg!(feature = "nexus") as u8
+            + cfg!(feature = "openvm") as u8
+            + cfg!(feature = "pico") as u8
+            + cfg!(feature = "risc0") as u8
+            + cfg!(feature = "sp1") as u8
+            + cfg!(feature = "zisk") as u8)
+            == 1,
+        "Exactly one zkVM backend feature must be enabled"
+    );
+};
```
han0110, line 21: Re https://github.com/eth-act/ere/pull/66#discussion_r2242879799 it works and this is shorter so I applied it.

crates/zkvm-interface/src/lib.rs
```diff
@@ -27,6 +28,7 @@ pub trait Compiler {
 
 /// ResourceType specifies what resource will be used to create the proofs.
 #[derive(Debug, Clone, Default)]
+#[cfg_attr(feature = "clap", derive(clap::Subcommand))]
```
han0110, line 31: Re https://github.com/eth-act/ere/pull/66#discussion_r2242906289 it seems much cleaner to do the derive so applied the suggestion.

crates/ere-cli/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+            resource,
+        } => execute(program_path, resource, input_path, report_path),
+        Commands::Verify {
+            program_path,
+            proof_path,
+        } => verify(program_path, proof_path),
+    }
+}
+
+fn compile(
```
han0110, line 117: Re https://github.com/eth-act/ere/pull/66#discussion_r2242911606 I think the macros seems a bit unnecessary so I remove it

Re https://github.com/eth-act/ere/pull/66#discussion_r2242917143 Previously there are some program sereizlied and some not, the reason is because some program is already `Vec<u8>` so I wanted to avoid double serialization, but it seems to be a footgun so I just serialize every program now.

crates/ere-cli/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+fn verify(program_path: PathBuf, proof_path: PathBuf) -> Result<(), Error> {
+    let zkvm = construct_zkvm(program_path, ProverResourceType::default())?;
+    let proof = fs::read(&proof_path)
+        .with_context(|| format!("Failed to read proof from {}", proof_path.display()))?;
+    zkvm.verify(&proof)
+        .with_context(|| "Failed to verify proof")?;
+    Ok(())
+}
+
+fn construct_zkvm(program_path: PathBuf, resource: ProverResourceType) -> Result<impl zkVM, Error> {
```
han0110, line 188: Re https://github.com/eth-act/ere/pull/66#discussion_r2242919308 I think somewhere has to do this once, currently it's done in `zkevm-benchmark-workload` like https://github.com/eth-act/zkevm-benchmark-workload/blob/5e44ca4dad7be4e531730a5eab8987efbc25dc38/crates/ere-hosts/src/main.rs#L206-L258, but I think in `ere` we could do this once so the dependents don't need to do it again and again.

PR #72, opened 2025-07-31
Fix `ere-jolt`

han0110:
Bugs found during working on #66.

- Added back mistakenly deleted `main.rs` of Jolt's guest in #68 
- Upgrade base dockerfile to use `ubuntu:24.04` for Jolt's compilation flow (it requires higher version of `GLIBC`)
- Install Jolt's toolchain when installing SDK to avoid installation while compiling guest

### DIFF
--- a/scripts/sdk_installers/install_jolt_sdk.sh
+++ b/scripts/sdk_installers/install_jolt_sdk.sh
@@ -34,6 +34,9 @@ ensure_tool_installed "cargo" "to build and install Rust packages"
 echo "Installing Jolt CLI from GitHub repository (a16z/jolt)..."
 cargo +nightly install --git https://github.com/a16z/jolt --force --bins jolt
 
+# Install Jolt's toolchain
+jolt install-toolchain
+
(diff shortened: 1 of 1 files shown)


jsign (approved):

docker/base/Dockerfile.base
```diff
@@ -1,9 +1,13 @@
...
 
-# Set DEBIAN_FRONTEND to noninteractive to avoid prompts during package installation
-ENV DEBIAN_FRONTEND=noninteractive
+# Free uid 1000 for later usage, see https://bugs.launchpad.net/cloud-images/+bug/2005129
+# for more details.
+RUN userdel -r ubuntu
+
+# Set DEBIAN_FRONTEND to noninteractive to avoid prompts during package
+# installation when building the image.
+ARG DEBIAN_FRONTEND=noninteractive
```
han0110, line 10: Use `ARG` instead of `ENV` here so it won't persist in the image.

kevaundray (approved):

PR #75, opened 2025-08-01
Add `ere-dockerized`

### LINKED ISSUE #14: Make Ere::compile use docker

han0110:
Turns out to be a bit huge still, but I think all the changes are needed by `ere-dockerized`

By commits it does:

1. Change `Compiler::compile` interface
    - Change `Compiler::compile(&Path, &Path)` to
  `Compiler::compile(&Path)`, now we only pass `guest_directory` and the compiler
  needs to figure out the workspace itself (only `EreDockerizedCompiler` needs to do that for mounting)
    - Change `Compiler::compile(&Path)` to
  `Compiler::compile(&self, &Path)`, so later `EreDockerizedCompiler`
  can be instantiated with a `enum ErezkVM` which does the dockerized
  compilation of instantiated zkVM.
    - Revert using docker for compilation of `ere-sp1`
  and `ere-risc0`, for dockerized compilation we should use `EreDockerizedCompiler`, this should also resolve #14 for all zkVM.
2. Add `ere-dockerized`
    - Add `enum ErezkVM` so we can select the zkVM we want to use
    - Builds image with the dockerfile `docker/cli/Dockerfile` with selected zkVM feature
    - Provides `EreDockerizedCompiler(ErezkVM)` that implements `Compiler` by executing docker command on the image
    - Provides `EreDockerizedzkVM { ErezkVM, .. }` that implements `zkVM` by executing docker command on the image
    - Also cleanup other dockerfiles so they don't do testing when build
3. Comment for zisk's script trick that should be added in #74

Ideally the usage of `ere-dockerized` should simplify `zkevm-benchmark-workload` a bit like https://github.com/han0110/zkevm-benchmark-workload/commit/0668b1f6745ef2d18e1c59d9005046e1e2a14f8c


crates/ere-dockerized/src/lib.rs
```diff
@@ -0,0 +1,451 @@
...
+        format!("ere-cli-{self}:{version}")
+    }
+
+    pub fn build_docker_image(&self) -> Result<(), CommonError> {
+        let workspace_dir = workspace_dir();
+
+        let force_rebuild = env::var("ERE_FORCE_REBUILD_DOCKER_IMAGE")
+            .ok()
+            .and_then(|s| s.parse::<bool>().ok())
+            .unwrap_or_default();
```
han0110, line 123: Re https://github.com/eth-act/ere/pull/66#discussion_r2242784471 indeed force rebuild seems safer.

docker/cli/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+WORKDIR /ere
+
+ARG ZKVM
+
+RUN cargo build --release --package ere-cli --bin ere-cli --features ${ZKVM} && \
+    cp /ere/target/release/ere-cli /ere/ere-cli && \
+    cargo clean && \
+    rm -rf $CARGO_HOME/registry/src $CARGO_HOME/registry/cache
+
+ENTRYPOINT ["/ere/ere-cli"]
```
han0110, line 16: Re https://github.com/eth-act/ere/pull/66#discussion_r2243022237 and https://github.com/eth-act/ere/pull/66#discussion_r2242788242 now it's a `Dockerfile` and looks much cleaner, previously there are more args so I decided to have a builder, but in the end it seems we only need to select `ZKVM`

crates/ere-dockerized/src/lib.rs
```diff
@@ -0,0 +1,503 @@
+//! # Ere Dockerized
```
han0110, line 1: Re  https://github.com/eth-act/ere/pull/66#discussion_r2242925511 add some comments and an example.

crates/ere-dockerized/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+
+[dependencies]
+bincode = "1.3"
+cargo_metadata = "0.19"
+serde = { version = "1.0", features = ["derive"] }
+tempfile = "3.3"
```
jsign, line 12: Nit: use workspace version?

crates/ere-dockerized/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+thiserror = "2"
+
+zkvm-interface = { workspace = true, features = ["clap"] }
+
+# Needed for Risc0 and OpenVM input serialization.
+bytemuck = "1.13"
+risc0-zkvm = { version = "^2.3.0", default-features = false }
```
jsign, line 19: Idea: could be better that we move all zkVM SDKs dependencies to the workspace `Cargo.toml`?

Before this PR the mental model for updating a zkVM SDK version was:
- Update `Cargo.toml` of the corresponding `ere-{{zkVM}}` crate.
- Update the pinned version in the corresponding zkVM Dockerfile.

This `ere-dockerized` breaks a bit that mental model, and I'm fearing we might forget to update some stuff. I can also agree that we can also have a mental model of doing Ctrl+F and look around though.

But I think if we move all SDKs to the workspace, and in each crate (being `ere-{{zkVM}}` or even this exception `ere-dockerized`), then we can have a simplified rule:
- Update in workspace `Cargo.toml`.
- Update pinned version in corresponding Dockerfile.

I think avoiding Ctrl+F is good since when reviewing PRs one have to trust the author did that work correctly, or reviewers have to clone and double-check.

No super strong opinions though -- just sharing an idea, but maybe have some drawbacks I could be missing.

crates/ere-dockerized/build.rs
```diff
@@ -0,0 +1,65 @@
+use build_utils::{detect_sdk_version, detect_self_crate_version};
+use std::{env, fs, path::Path};
+
+fn main() {
+    generate_ere_version();
```
jsign, line 5: Thought not for this PR: maybe at some point it would be good to also append the git hash, since some people could be using `master` and could be useful?

crates/ere-dockerized/build.rs
```diff
@@ -0,0 +1,65 @@
...
+        "jolt-sdk",
+        "nexus-sdk",
+        "openvm-sdk",
+        "pico-sdk",
+        "risc0-zkvm",
+        "sp1-sdk",
+    ]
+    .map(detect_sdk_version);
+
+    // FIXME: ZisK doens't depend on SDK yet, so we hardcode the version here,
```
jsign, line 39:
```suggestion
    // FIXME: ZisK doesn't depend on SDK yet, so we hardcode the version here,
```

crates/ere-dockerized/src/input.rs
```diff
@@ -0,0 +1,64 @@
...
+        match self {
+            // Issue for tracking: https://github.com/eth-act/ere/issues/4.
+            Self::Jolt => unimplemented!(),
+            // Issue for tracking: https://github.com/eth-act/ere/issues/63.
+            Self::Nexus => unimplemented!(),
+            // FIXME: Instead of using `openvm::serde::to_vec`, we use Risc0's
+            //        serializer, because OpenVM uses the same one, to avoid the
+            //        duplicated extern symbol they export.
+            //        It'd be better to have each zkvm provides their
+            //        lightweight serde crate.
```
jsign, line 19: Worth also creating an issue?

crates/ere-dockerized/src/lib.rs
jsign: I think the safer option is the one you did initially :smile::
- Original take: If no env is provided, it will always rebuild
- New take: If not env is provided, it won't rebuild (i.e., use cached).

I'm fine with both -- just double checking here if you really intended to do the safer one?

crates/ere-dockerized/src/lib.rs
```diff
@@ -0,0 +1,512 @@
...
+
+/// Wrapper for serialized program.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct SerializedProgram(Vec<u8>);
+
+impl Compiler for EreDockerizedCompiler {
+    type Error = CompileError;
+    type Program = SerializedProgram;
+
+    fn compile(&self, guest_directory: &Path) -> Result<Self::Program, Self::Error> {
```
jsign, line 205: How would the workload repo use this now?

For example, `guest_directory`would be [this path](https://github.com/eth-act/zkevm-benchmark-workload/tree/master/ere-guests/stateless-validator/sp1)? How it decides to mount the [underlying workspace](https://github.com/eth-act/zkevm-benchmark-workload/tree/master/ere-guests)?

docker/zisk/Dockerfile
```diff
@@ -37,14 +38,21 @@ RUN wget https://developer.download.nvidia.com/compute/cuda/repos/$(. /etc/os-re
...
-# This script installs the 'zisk' Rust toolchain and cargo-zisk.
-# TODO: Download the proving key if the CI runner has enough disk space.
-RUN SETUP_KEY=verify /tmp/install_zisk_sdk.sh && rm /tmp/install_zisk_sdk.sh # Clean up the script
+# This script installs the 'zisk' Rust toolchain and `cargo-zisk`
+#
+# If argument `CI` is set, we only install verifying key, this is used by github
+# CI runner which only has small disk space (proving key requires ~35 GB).
+RUN if [ -n "$CI" ]; then export SETUP_KEY=verify; fi && \
+    /tmp/install_zisk_sdk.sh && \
+    rm /tmp/install_zisk_sdk.sh # Clean up the script
```
jsign, line 55: Asking to understand more:
1. These verifying and proving keys are related to aggregation circuits?
2. What's the consequence of only installing one key but not the other?
3. Maybe with the devops GH runners we can remove this since they have 100GiB?

Sorry some are basic questions :)

jsign (approved):
LGTM! I left some comments/questions.

crates/ere-dockerized/Cargo.toml
han0110: Yes this makes sense, I was actually planning to move all dependencies to workspace if there is no explicit version conflict, ideally we only need to touch 3 places if we do so:

- `<repo>/Cargo.toml`
- `docker/{zkvm}/Dockerfile`
- `tests/{zkvm}/**/Cargo.toml`

Is it okay to do so in another PR? I was afraid the diff gets bigger so haven't done it yet.

jsign: Yes yes, was mostly an idea for another PR if makes sense.

crates/ere-dockerized/src/lib.rs
han0110: Ah I think we could keep it similar with the original one, but make it an instance method like `fn compile(&self, workspace_directory: &Path, guest_relative: &Path)`, so `EreDockerizedCompiler(ErezkVM)` could work.

docker/zisk/Dockerfile
han0110: > Reg 1. I think I was mostly thinking on proving and verifying keys that usually depend on the program you want to prove.

It indeed depends on their zkVM circuit and the aggregation circuit I think, it does some pre-processing when installation that makes the proving faster but takes huge disk space. Tho it doesn't depend on the guest program because guest program could be seen as a private input to the zkVM circuit.

crates/ere-dockerized/src/lib.rs
jsign: Makes sense, so then `EreDockerizedCompiler` won't impl `Compiler`?

docker/zisk/Dockerfile
jsign: Ah sorry, I think I was sticking to other non-zkVM thinking where guest programs were custom -- in this case the VM is the program and the "guest programs" are inputs yes.

crates/ere-dockerized/src/input.rs
kevaundray: >   It'd be better to have each zkvm provides their lightweight serde crate.

This seems like a good suggestion, can we create an issue so we remember to ask them in the future to do this?

crates/ere-dockerized/src/lib.rs
jsign: BTW, feel free to merge this PR as is -- we can continue the convo in other PR.

han0110: So in the end I did this https://github.com/eth-act/ere/pull/75/commits/58575d91d6e1d8dc1deb216e537339180761976e, I think this should be flexible enough to support any kinds of layout (actually similar as previous, but now we have 2 absolute path `mounting_directory` and `guest_directory` and require the latter is sub-dir of the former).

PR #79, opened 2025-08-07
Cleanup some todos in SP1 dockerfile

han0110:
Also noticed that SP1 doesn't really use the env `SP1_DIR` in their `cargo-prove` CLI, when it installs the toolchain [it always install to `$HOME/.sp1`](https://github.com/succinctlabs/sp1/blob/v5.1.0/crates/cli/src/commands/install_toolchain.rs#L61).

So not sure if we should keep the `SP1_DIR` a argument or not.

### DIFF
--- a/scripts/sdk_installers/install_sp1_sdk.sh
+++ b/scripts/sdk_installers/install_sp1_sdk.sh
@@ -14,29 +14,27 @@ if ! command -v bash &> /dev/null; then # sp1up script uses bash
 fi
 
 # Define default homes if not set, useful for Docker context
-DEFAULT_SP1UP_HOME="${HOME}/.sp1up"
-DEFAULT_SP1_HOME="${HOME}/.sp1"
+DEFAULT_SP1_DIR="${HOME}/.sp1"
 
 # Use existing ENV var or default. Docker ENV will make these available.
 # For local use, user might need to add these to their .bashrc/.zshrc
-export SP1UP_HOME="${SP1UP_HOME:-${DEFAULT_SP1UP_HOME}}"
-export SP1_HOME="${SP1_HOME:-${DEFAULT_SP1_HOME}}"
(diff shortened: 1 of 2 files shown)


scripts/sdk_installers/install_sp1_sdk.sh
```diff
@@ -14,29 +14,27 @@ if ! command -v bash &> /dev/null; then # sp1up script uses bash
...
 
 # Use existing ENV var or default. Docker ENV will make these available.
 # For local use, user might need to add these to their .bashrc/.zshrc
-export SP1UP_HOME="${SP1UP_HOME:-${DEFAULT_SP1UP_HOME}}"
-export SP1_HOME="${SP1_HOME:-${DEFAULT_SP1_HOME}}"
+export SP1_DIR="${SP1_DIR:-${DEFAULT_SP1_DIR}}"
 
 # Run sp1up installer script
-curl -L https://sp1up.succinct.xyz | bash -s -- --yes 
+curl -L https://sp1up.succinct.xyz | bash
```
han0110, line 24: No interactoin in sp1up so removing the `--yes`

scripts/sdk_installers/install_sp1_sdk.sh
```diff
@@ -14,29 +14,27 @@ if ! command -v bash &> /dev/null; then # sp1up script uses bash
...
 
 # Add sp1up and sp1 binaries to PATH for this script's execution context
 # and for subsequent commands if this script is sourced.
-export PATH="${SP1UP_HOME}/bin:${SP1_HOME}/bin:$PATH"
+export PATH="${SP1_DIR}/bin:$PATH"
 
-export SDK_VERSION="${SP1UP_SDK_INSTALL_VERSION:-latest}"
+export SP1_VERSION="${SP1_VERSION:-latest}"
```
han0110, line 30: Rename it to be same

PR #80, opened 2025-08-08
Add `InputItem::SerializedObject`

han0110:
This is required for using dockerized `risc0` (`ere-dockerized` calling `ere-cli` with feature `risc0`).

In `risc0` there are multiple way to write the input:
- `write::<T: Serialize>(obj: T)` - It serializes the object with its own serializer, used to serialize `InputItem::Object`
- `write_frame(bytes: Vec<u8>)` - It serializes it by `[bytes.len().as_le_bytes(), bytes].concat()`, use to serialize `InputItem::Bytes`
- `write_slice(bytes: Vec<u8>)` - It doesn't do extra serialization and just append the bytes to its input, not used in ere currently

Currently `ere-dockerized` needs to serialize the `InputItem::Object` to be able to pass it to `ere-cli` by file, this works for most zkVMs that doesn't do extra serialization on the `InputItem::Bytes`, but doesn't work for `risc0`.

For this PR add an extra variant `InputItem::SerializedObject` for `risc0` to identify that it's already serialized in its own serializer, and shuold be appended to its input directly by calling `write_slice`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2327,6 +2327,7 @@ dependencies = [
  "bincode",
  "build-utils",
  "bytemuck",
+ "ere-cli",
  "risc0-zkvm",
  "serde",
  "tempfile",
@@ -2387,6 +2388,7 @@ dependencies = [
  "bincode",
  "build-utils",
  "pico-sdk",
+ "pico-vm",
  "thiserror 2.0.12",
  "zkvm-interface",
 ]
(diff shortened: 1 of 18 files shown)


jsign (commented):
I think makes sense, but I'm left with a feeling of being a bit unfortunate.

Maybe worth getting extra feedback on this one from @kevaundray .

kevaundray:
> Leaving the approval if you want to merge. We can keep marinating if we can improve on simplifying this weirdish new variant in the public API in a future PR or something.
> 
> Not sure it is worth blocking progress because of this, we can maybe figure out what to do in the near future.

Yeah, I think we'll need time to think about whats the best API here -- in the meanintime, we can merge this as you said and open up an issue so we come back to it in the future

PR #82, opened 2025-08-13
Add crate `test-utils`

### LINKED ISSUE #15: Remove duplication in tests
Most of the tests between zkVMs look similar. Given we have an interface, we can use that to DRY with some of the tests

han0110:
This PR aims to add more complicated input type to testing guests, and simplify the testing guest structure a bit.

The `test-utils` provides:

- Serde types for guest/host to use
- Helper function to get path to guest directory
- Helper function to get sample inputs

Currently only `ere-sp1` is updated, if it looks okay, I'll apply it to all the other zkvm crates and fully resolve #15.

### DIFF
--- a/tests/sp1/basic/src/main.rs
+++ b/tests/sp1/basic/src/main.rs
@@ -10,8 +10,8 @@ pub fn main() {
     assert_eq!(String::from_utf8_lossy(&bytes), "Hello world");
 
     // Read `BasicStruct`.
-    let basic_input = sp1_zkvm::io::read::<BasicStruct>();
-    let output = basic_input.output();
+    let basic_struct = sp1_zkvm::io::read::<BasicStruct>();
+    let output = basic_struct.output();
 
     // Write `output`
     sp1_zkvm::io::commit(&output);


crates/ere-sp1/src/compile.rs
```diff
@@ -95,50 +95,21 @@ pub fn compile(guest_directory: &Path) -> Result<Vec<u8>, CompileError> {
...
-            }
-            Err(err) => {
-                panic!("compile failed for dedicated guest: {err}");
-            }
-        }
+    fn test_compile() {
+        let guest_directory = testing_guest_directory("sp1", "basic");
+        let elf_bytes = compile(&guest_directory).unwrap();
+        assert!(!elf_bytes.is_empty(), "ELF bytes should not be empty.");
     }
```
jsign, line 107: Q: Maybe we can remove this and only leave the one below?

crates/ere-sp1/src/lib.rs
```diff
@@ -225,127 +225,62 @@ fn serialize_inputs(stdin: &mut SP1Stdin, inputs: &Input) {
...
-            .join("prove")
-            .join("basic")
-            .canonicalize()
-            .expect("Failed to find or canonicalize test guest program at <CARGO_WORKSPACE_DIR>/tests/execute/sp1")
-    }
-
-    fn get_compiled_test_sp1_elf_for_prove() -> Result<Vec<u8>, SP1Error> {
-        let test_guest_path = get_prove_test_guest_program_path();
-        RV32_IM_SUCCINCT_ZKVM_ELF.compile(&test_guest_path)
+        run_zkvm_execute_invalid_inputs(&zkvm, &empty_inputs);
```
jsign, line 264: Q: Maybe worth also some non-empty imput that isn't the expected `Hello world`?

jsign: Q2/Nit: might be useful to remove `run_zkvm_execute_invalid_inputs` and only have the normal one and do the `.expect` here?

crates/test-utils/src/host.rs
```diff
@@ -0,0 +1,56 @@
...
+pub fn testing_guest_directory(zkvm_name: &str, program: &str) -> PathBuf {
+    workspace().join("tests").join(zkvm_name).join(program)
+}
+
+pub fn run_zkvm_execute(zkvm: &impl zkVM, inputs: &Input) {
+    let _report = zkvm
+        .execute(inputs)
+        .expect("execute should not fail with valid input");
+
+    // TODO: Check output are expected.
```
jsign, line 21: Might worth creating an issue to include program outputs into `ProgramExecutionReport` and `ProgramProvingReport`?

jsign (approved):
LGTM! Left some comments for your consideration.

Also, I think this might close https://github.com/eth-act/ere/issues/15 ?

crates/test-utils/src/host.rs
han0110: Sure, created here https://github.com/eth-act/ere/issues/87

But if we only include the program outpus in report, it seems verifier still couldn't get it, so perhaps we also need to update the `verify` to return the `Output` (or `ProgramVerifyingReport` that contains `Output`)?

crates/ere-sp1/src/lib.rs
han0110: > Q: Maybe worth also some non-empty imput that isn't the expected Hello world?

Sounds good

> Q2/Nit: might be useful to remove run_zkvm_execute_invalid_inputs and only have the normal one and do the .expect here?

Yeah we could do that, but then we will duplicate a lot of `.expect(msg)` in tests. Do you think we could just use `unwrap()`, or it's fine to duplicate the expect message?

PR #88, opened 2025-08-14
Use `test-utils` for all zkvms

### LINKED ISSUE #15: Remove duplication in tests

han0110:
Fully resolves #15 

Some zkvm have different (simpler) tests than others, and only `compile` is tested:
- Jolt - It doesn't support private input yet
- Nexus - Its target `riscv32i-unknown-none-elf` doesn't compile crate `serde`


crates/ere-zisk/src/lib.rs
```diff
@@ -284,13 +295,17 @@ impl zkVM for EreZisk {
...
 fn serialize_inputs(inputs: &Input) -> Result<Vec<u8>, bincode::Error> {
     inputs.iter().try_fold(Vec::new(), |mut acc, item| {
         match item {
-            InputItem::Object(obj) => {
-                bincode::serialize_into(&mut acc, &**obj)?;
-            }
-            InputItem::SerializedObject(bytes) | InputItem::Bytes(bytes) => acc.extend(bytes),
+            InputItem::Object(obj) => bincode::serialize_into(&mut acc, &**obj)?,
+            InputItem::SerializedObject(bytes) => acc.extend(bytes),
+            InputItem::Bytes(bytes) => bincode::serialize_into(&mut acc, bytes)?,
```
han0110, line 308: We need to treat the bytes as object and serialize it as well, otherwise the deserializer doesn't know the length of the bytes.

`zisk`'s guest program in `zkevm-benchmark-workload` also needs to be updated.

crates/ere-zisk/src/lib.rs
```diff
@@ -24,6 +25,10 @@ include!(concat!(env!("OUT_DIR"), "/name_and_sdk_version.rs"));
 mod compile;
 mod error;
 
+/// It panics if `EreZisk::prove` is called concurrently, so we need a lock here
+/// to avoid that.
+static PROVE_LOCK: Mutex<()> = Mutex::new(());
```
han0110, line 30: Add a lock here so we can run the test without needing the flag `--test-threads=1` or worrying about the prove method gets called concurrently.

jsign (approved):

PR #91, opened 2025-08-15
Ci/test more and push image

han0110:
- Build the base zkvm image and push to ghcr, only rebuild when `Cargo.toml` version changes
- Test the crate `ere-{zkvm}` via docker base zkvm image
- Test `ere-dockerized`

### DIFF
--- a/.github/workflows/test-zkvm-jolt.yml
+++ b/.github/workflows/test-zkvm-jolt.yml
@@ -9,6 +9,9 @@ on:
 jobs:
   test:
     uses: ./.github/workflows/test-zkvm.yml
+    permissions:
+      contents: read
+      packages: write
     with:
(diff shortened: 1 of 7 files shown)


han0110:
The registry seems to be unusabe for this repo so the image building fails. @jsign Do you know how to enable that, or anything extra I need to do in the CI script?

jsign:
> The registry seems to be unusabe for this repo so the image building fails. @jsign Do you know how to enable that, or anything extra I need to do in the CI script?

@han0110, I think [these lines](https://github.com/eth-act/zkevm-benchmark-workload/blob/13f882a6ae46bd9bef8fb116c76796fe19824809/.github/workflows/docker-build.yml#L11-L13) are missing.

PR #100, opened 2025-08-20
Risc0 check receipt variant

han0110:
- Add `Risc0ProofWithPublicValues` to make sure the inner receipt is always `SuccinctReceipt`
- Add `zkVMError::other` to reduce the need to call `Box::new` or `.into()` too much


crates/ere-risc0/src/lib.rs
```diff
@@ -101,17 +143,20 @@ impl zkVM for EreRisc0 {
             }
         };
 
-        let encoded = borsh::to_vec(&receipt).map_err(|err| zkVMError::Other(Box::new(err)))?;
-        Ok((encoded, ProgramProvingReport::new(proving_time)))
+        let proof = borsh::to_vec(&Risc0ProofWithPublicValues::try_from(receipt)?)
+            .map_err(zkVMError::other)?;
```
jsign, line 147: Is the `prove_with_opts` to indicate which kind of Receipt we want lost for the multi-gpu proving? (I'm wondering why we need to do this checking after the fact compared with intentionally always asking for `Succinct` type)

jsign (approved):

crates/ere-risc0/src/lib.rs
han0110: hmm yeah, you are right, we should be able to just unwrap here, will update the make thing simpler.

han0110: I think by default the bento prover creates succinct proof.

PR #103, opened 2025-08-21
Upgrade `risc0` to `3.0.1` and use the new actor system prover

### LINKED ISSUE #81: Risc0 GPU proving takes extra time for the first proof
Currently Risc0's bento proving stack, it takes ~16 extra minutes to generate the first proof (5090).

Specifically the following 3 operations takes much longer time:
- Proving segment - ~130 sec (`grep -m 2 -P "(Starting proof of idx: .* - 0|Completed proof: .* - 0)" {log}`)
- Proving lift - ~8 sec (`grep -m 2 -P "(Proving lift: .*|Proving lift finished: .*)" {log}`)
- Proving keccak - ~810 sec (`grep -m 2 -P "(Keccak proving .*|Completed keccak proving .*)" {log}`)

han0110:
- Upgrade `risc0` to `3.0.1` and use the new prover that spawn the workers (actors) directly instead of using docker compose
    - Much less complicated than the docker compose approach
    - Now we need to confiugre `RISC0_DEFAULT_PROVER_NUM_GPUS` to enable multi-gpu proving, instead of using `CUDA_VISIBLE_DEVICES`
- In Risc0's SDK installation script, now we need to build the `r0vm` from source with the `cuda` feature enabled (previously this is done at runtime by build docker image)
    - Also skip the GPU related commands in CI env to not slowdown it

Resolves #81. Previously the issue that first proving takes longer is becasue [their nvcc flag is not generating native code for 5090 by default](https://github.com/risc0/risc0/blob/96eef1c3c9c6dde65dcc2fc86791a4358fcc8e4d/bento/dockerfiles/agent.dockerfile#L36), now we don't use their image so it's no longer an issue anymore.


Cargo.toml
```diff
@@ -70,9 +70,8 @@ pico-vm = { git = "https://github.com/brevis-network/pico.git", tag = "v1.1.4" }
 pico-sdk = { git = "https://github.com/brevis-network/pico.git", tag = "v1.1.4" }
 
 # Risc0 dependencies
-risc0-build = "2.3.1"
-risc0-zkvm = { version = "2.3.1", default-features = false }
-bonsai-sdk = { version = "1.4.0" }
+risc0-build = { git = "https://github.com/risc0/risc0.git", tag = "v3.0.0" }
+risc0-zkvm = { git = "https://github.com/risc0/risc0.git", tag = "v3.0.0", default-features = false }
```
han0110, line 74: Update to `crates.io` when it's released (it's once released but not sure why yanked https://crates.io/crates/risc0_zkvm/3.0.0)

scripts/sdk_installers/install_risc0_sdk.sh
```diff
@@ -82,4 +86,23 @@ cargo risczero --version || (echo "Error: cargo risczero command failed!" >&2 &&
...
+    CARGO_HOME="${CARGO_HOME:-$HOME/.cargo}"
+    RISC0_BIN_DIR="$HOME/.risc0/extensions/v$RISC0_VERSION-cargo-risczero-x86_64-unknown-linux-gnu"
+
+    TEMP_DIR=$(mktemp -d)
+    git clone https://github.com/risc0/risc0.git --depth 1 --branch "v$RISC0_VERSION" "$TEMP_DIR/risc0"
+    cd "$TEMP_DIR/risc0"
+
+    # Build with feature `cuda`
+    RUSTFLAGS="-C target-cpu=native" cargo build --release --features cuda --bin r0vm
+    # Copy the binary the the same directory with `cargo-risczero` and `r0vm`
```
jsign, line 102:
```suggestion
    # Copy the binary to the same directory with `cargo-risczero` and `r0vm`
```

jsign (approved):
LGTM! Great job -- this is quite deep into Risc0 internals :)

PR #108, opened 2025-08-27
Upgrade sp1 to 5.2.1

SuccinctPaul:
Install golang in docker base as most zkvm use gnark to implement the snark circuit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -63,6 +63,165 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
+name = "alloy-consensus"
+version = "1.0.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ad451f9a70c341d951bca4e811d74dbe1e193897acd17e9dbac1353698cc430b"
+dependencies = [
+ "alloy-eips",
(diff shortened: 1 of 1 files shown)


docker/base/Dockerfile.base
```diff
@@ -52,6 +53,8 @@ WORKDIR /app
 # TODO: Default to the non-root user?
 # USER ${USERNAME}
 
+# Verify go installation went well
+RUN go version
```
han0110, line 57:
```suggestion
```

docker/base/Dockerfile.base
```diff
@@ -24,6 +24,7 @@ RUN apt-get update && \
     ca-certificates \
     openssl \
     libssl-dev \
+    golang \
```
han0110, line 27: I think for now we are not focusing on small proof that can be verified on chain, so I think it'd be better to not install go that increases the image size.

```suggestion
```

crates/ere-sp1/src/lib.rs
```diff
@@ -21,6 +21,7 @@ mod compile_stock_rust;
 mod error;
 use error::{ExecuteError, ProveError, SP1Error, VerifyError};
 
+#[allow(clippy::large_enum_variant)]
```
han0110, line 24: Any reason to add this?

han0110 (commented):
Thanks for the PR, the `SP1_VERSION="v5.1.0"` in `docker/sp1/Dockerfile` also needs to be updated.

PR #109, opened 2025-08-27
Implement returning of public values for other zkVMs

han0110:
This PR by commits:

- Implement returning of public values for `risc0`
- Implement returning of public values for `openvm`
    - Because it doesn't have native serialization for committing output, so it returns error in `zkVM::deserialize_from`
    - Implement a `OutputHashedIo` for `openvm` to check the hashed public values are as expected
    - Upgrade the version to latest release candidate
- Implement returning of public values for `zisk`, using the same logic as `openvm`
- Implement returning of public values for `pico`, but `verify` is left as todo as is.

### DIFF
--- a/crates/ere-openvm/src/lib.rs
+++ b/crates/ere-openvm/src/lib.rs
@@ -177,12 +177,7 @@ impl zkVM for EreOpenVM {
             }))?;
         }
 
-        let public_values = proof
-            .user_public_values
-            .iter()
-            .map(|v| u8::try_from(v.as_canonical_u32()).ok())
-            .collect::<Option<_>>()
-            .ok_or_else(|| OpenVMError::from(ProveError::from(CommonError::InvalidPublicValue)))?;
+        let public_values = extract_public_values(&proof.user_public_values)?;
         let proof_bytes = proof
             .encode_to_vec()
             .map_err(|e| OpenVMError::from(ProveError::SerializeProof(e)))?;
@@ -201,12 +196,7 @@ impl zkVM for EreOpenVM {
         Sdk::verify_proof(&self.agg_vk, self.app_commit, &proof)
             .map_err(|e| OpenVMError::Verify(VerifyError::Verify(e)))?;
 
-        let public_values = proof
-            .user_public_values
-            .iter()
-            .map(|v| u8::try_from(v.as_canonical_u32()).ok())
-            .collect::<Option<_>>()
-            .ok_or_else(|| OpenVMError::from(VerifyError::from(CommonError::InvalidPublicValue)))?;
+        let public_values = extract_public_values(&proof.user_public_values)?;
 
         Ok(public_values)
     }
@@ -235,6 +225,18 @@ fn serialize_inputs(stdin: &mut StdIn, inputs: &Input) {
             }
         }
     }
+}
+
+/// Extract public values in bytes from field elements.
+///
+/// The public values revealed in guest program will be flatten into `Vec<u8>`
+/// then converted to field elements `Vec<F>`, so here we try to downcast it.
+fn extract_public_values(user_public_values: &[F]) -> Result<Vec<u8>, CommonError> {
+    user_public_values
+        .iter()
+        .map(|v| u8::try_from(v.as_canonical_u32()).ok())
+        .collect::<Option<_>>()
+        .ok_or(CommonError::InvalidPublicValue)
 }
 
 #[cfg(test)]


crates/ere-risc0/src/lib.rs
```diff
@@ -232,8 +232,29 @@ impl zkVM for EreRisc0 {
...
+                self.0
+                    .read_exact(bytes)
+                    .map_err(risc0_zkvm::serde::Error::custom)?;
+                self.0
+                    .read_exact(&mut padded_bytes)
+                    .map_err(risc0_zkvm::serde::Error::custom)
+            }
+        }
+
+        T::deserialize(&mut Deserializer::new(WordReadAdapter(reader))).map_err(zkVMError::other)
```
jsign, line 257: I was looking at [the docs](https://dev.risczero.com/api/zkvm/tutorials/io#reading-public-data-in-the-host), and they have this [decode()](https://github.com/risc0/risc0/blob/c1ea88f36839b889a15de419e1190ae34c2e0932/risc0/zkvm/src/receipt.rs#L285).. but looks like this is assuming the program only commited as ingle value, and here we want to support multiple commits?

crates/ere-zisk/src/lib.rs
```diff
@@ -352,6 +358,42 @@ fn serialize_inputs(inputs: &Input) -> Result<Vec<u8>, bincode::Error> {
...
+    // The first 4 elements of public values should be ROM digest.
+    let rom_digest = public_values[..4].try_into().unwrap();
+
+    // The 5-th element should be the number of user set public values.
+    let num_user_public_values = public_values[4] as usize;
+
+    // The rest elements should be user set public values and should be `u32`.
+    let public_values = public_values[5..]
+        .iter()
+        .map(|v| Some(u32::try_from(*v).ok()?.to_le_bytes()))
```
jsign, line 386: Was wondering why `u32`, and ended up [here](https://0xpolygonhermez.github.io/zisk/getting_started/writing_programs.html#inputoutput-data). I guess makes sense.

Let's hope eventually they include this function in their SDK, since they might change this in the future and we might have the wrong interpretation of how things are encoded. (Not much we can do for now!)

jsign (approved):
LGTM! Left some comments for your consideration.

crates/ere-openvm/src/lib.rs
```diff
@@ -180,13 +177,16 @@ impl zkVM for EreOpenVM {
             }))?;
         }
 
+        let public_values = proof
+            .user_public_values
+            .iter()
+            .map(|v| u8::try_from(v.as_canonical_u32()).ok())
```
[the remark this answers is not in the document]
han0110: Ah forget to add some note here, the reason to convert to `u8` is because the [public values are extracted into `Vec<u8>`](https://github.com/openvm-org/openvm/blob/v1.4.0-rc.8/crates/vm/src/system/memory/merkle/public_values.rs#L168-L187), but for some reason in proof [it's converted to field elements](https://github.com/openvm-org/openvm/blob/v1.4.0-rc.8/crates/vm/src/system/memory/merkle/public_values.rs#L66-L69), that's why we could convert it back here.

> This is because the only way to commit outputs is with reveal_32 which encodes 32 bytes into 32 31-bit field elements which are u8. (Asking to double check).

For some reason the revealed values are converted to `Vec<u8>` then to field 31-bits field elements in their proving process, but not sure why for the guest program it doesn't provide commit to `&[u8]`, even in `reveal_bytes32` it calls . I'm guessing it's mainly due to riscv32 reason.

> Is there other way to commit to a field element, thus this invariant doesn't hold? (I guess the answer is "no", but maybe someone want's to use all field bits)

For the user public values I think it'd always be extracted as `Vec<u8>`, and from their API it seems no way to do this.

crates/ere-openvm/src/lib.rs
```diff
@@ -215,8 +219,10 @@ impl zkVM for EreOpenVM {
         SDK_VERSION
     }
 
-    fn deserialize_from<R: Read, T: DeserializeOwned>(&self, _reader: R) -> Result<T, zkVMError> {
-        todo!()
+    fn deserialize_from<R: Read, T: DeserializeOwned>(&self, _: R) -> Result<T, zkVMError> {
```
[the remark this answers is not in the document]
han0110: Ah good point. Tho it might makes the user API more complicated, but it should reduce the surprise at runtime.

crates/ere-risc0/src/lib.rs
han0110: Yes, if the guest program only `commit` single thing, we can use the `decode` function, but to support multiple `commit`, we need to have this adapter.

crates/ere-zisk/src/lib.rs
han0110: Yes, their guest program only have the `set_output(idx: usize, val: u32)` option to commit public values, so here we downcast it from `u64`.

> Let's hope eventually they include this function in their SDK, since they might change this in the future and we might have the wrong interpretation of how things are encoded.

Make sense!

PR #112, opened 2025-08-28
zkVM output for `ere-dockerized`

han0110:
This PR:

- Implements `zkVM::deserialize_from` for `EreDockerized` (for `risc0` we share the same code without adding dependency to keep `ere-dockerized` lightweight)
- Move the output check into `test-utils`
- Use `todo!()` to mark that it should be figured out but not yet done (`jolt` and `nexus`), and use `unimplemented!()` to mark that it's platform restriction so will never be implemented (`openvm` and `zisk`)

### DIFF
--- a/crates/ere-dockerized/src/lib.rs
+++ b/crates/ere-dockerized/src/lib.rs
@@ -513,7 +513,7 @@ mod test {
 
         let zkvm = EreDockerizedzkVM::new(zkvm, program, ProverResourceType::Cpu).unwrap();
 
-        let io = BasicProgramIo::valid();
+        let io = BasicProgramIo::valid().into_output_hashed_io();
         run_zkvm_execute(&zkvm, &io);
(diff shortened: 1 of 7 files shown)


crates/ere-dockerized/src/lib.rs
```diff
@@ -215,6 +216,10 @@ impl EreDockerizedCompiler {
             mount_directory: mount_directory.as_ref().to_path_buf(),
         }
     }
+
+    pub fn zkvm(&self) -> ErezkVM {
+        self.zkvm
+    }
```
han0110, line 222: Adding this so `zkevm-benchmark-workload` can match the `zkvm` and decide to deserialize the output and check (`sp1` and `risc0`), or hash the expected output and check (`openvm` and `zisk`).

jsign (approved):

PR #113, opened 2025-08-28
Add GPU proving support in `ere-openvm`

han0110:
- Update `openvm` to `v1.4.0`.
- Add `cuda` feature and enabled by default in `ere-openvm`, and add support of `ProverResourceType::Gpu`
- Add feature flag `cuda` in `ere-cli` as well
- In CI add an boolean option `default_features` to conditional enable/disable default features, which includes `cuda`, to avoid needing cuda to run clippy/test.

I thin eventually we need to handle the cuda building thing on CI, otherwise I feel the building process will get more and more complicated :(

### DIFF
--- a/.github/workflows/test-zkvm.yml
+++ b/.github/workflows/test-zkvm.yml
@@ -17,6 +17,12 @@ on:
         required: false
         type: boolean
         default: true
+      # Currently only needed by those zkvms that need Cuda during building process by default
+      default_features:
+        description: 'Cargo clippy/build with default features or not'
+        required: false
+        type: boolean
+        default: true
       # Remove when we use larger runners, currently only needed to skip some zisk test
       test_options:
         description: 'Cargo test options when testing via Docker'
@@ -45,12 +51,14 @@ jobs:
         run: |
           cargo clippy --all-targets \
             --package ere-${{ inputs.zkvm }} \
+            ${{ inputs.default_features && '' || '--no-default-features' }} \
             -- -D warnings
 
       - name: Run cargo clippy for ere-cli with feature ${{ inputs.zkvm }}
         run: |
           cargo clippy --all-targets \
             --package ere-cli \
+            ${{ inputs.default_features && '' || '--no-default-features' }} \
             --features cli,${{ inputs.zkvm }} \
             -- -D warnings
 
@@ -64,6 +72,7 @@ jobs:
       image_version: ${{ steps.inspect_image.outputs.image_version }}
       base_image_tag: ${{ steps.inspect_image.outputs.base_image_tag }}
       base_zkvm_image_tag: ${{ steps.inspect_image.outputs.base_zkvm_image_tag }}
+      cli_zkvm_image_tag: ${{ steps.inspect_image.outputs.cli_zkvm_image_tag }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -86,10 +95,12 @@ jobs:
           IMAGE_VERSION="$ZKVM_CRATE_VERSION-$GIT_REV"
           BASE_IMAGE_TAG="ghcr.io/${{ github.repository }}/ere-base:$IMAGE_VERSION"
           BASE_ZKVM_IMAGE_TAG="ghcr.io/${{ github.repository }}/ere-base-${{ inputs.zkvm }}:$IMAGE_VERSION"
+          CLI_ZKVM_IMAGE_TAG="ghcr.io/${{ github.repository }}/ere-cli-${{ inputs.zkvm }}:$IMAGE_VERSION"
 
           echo "image_version=$IMAGE_VERSION" >> $GITHUB_OUTPUT
           echo "base_image_tag=$BASE_IMAGE_TAG" >> $GITHUB_OUTPUT
           echo "base_zkvm_image_tag=$BASE_ZKVM_IMAGE_TAG" >> $GITHUB_OUTPUT
+          echo "cli_zkvm_image_tag=$CLI_ZKVM_IMAGE_TAG" >> $GITHUB_OUTPUT
 
       - name: Build ere-base image
         uses: docker/build-push-action@v6
@@ -107,8 +118,8 @@ jobs:
           push: true
           tags: ${{ steps.inspect_image.outputs.base_zkvm_image_tag }}
           build-args: |
-            CI=1
             BASE_IMAGE_TAG=${{ steps.inspect_image.outputs.base_image_tag }}
+            CI=1
 
   test_via_docker:
     name: Test via Docker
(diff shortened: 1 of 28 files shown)


.github/workflows/test-zkvm.yml
```diff
@@ -86,10 +95,12 @@ jobs:
           IMAGE_VERSION="$ZKVM_CRATE_VERSION-$GIT_REV"
           BASE_IMAGE_TAG="ghcr.io/${{ github.repository }}/ere-base:$IMAGE_VERSION"
           BASE_ZKVM_IMAGE_TAG="ghcr.io/${{ github.repository }}/ere-base-${{ inputs.zkvm }}:$IMAGE_VERSION"
+          CLI_ZKVM_IMAGE_TAG="ghcr.io/${{ github.repository }}/ere-cli-${{ inputs.zkvm }}:$IMAGE_VERSION"
```
jsign, line 98: Nice, new images to exploit in workload :smile:

jsign:
> I thin eventually we need to handle the cuda building thing on CI, otherwise I feel the building process will get more and more complicated :(

By this, you mean requiring a GH runner with a GPU? Or having a more beefy machine to support installing and compiling kernels in a reasonable amount of time? I think if it is the latter, we can ask the devops team to have a more powerful runner than the most powerful documented one. In the docs they say this is possible, but has to be asked explicitly. 

If that could help, I can reach out to the devops team to explore our options.

crates/ere-dockerized/src/lib.rs
```diff
@@ -161,8 +161,15 @@ impl ErezkVM {
                 )
                 .tag(self.base_zkvm_tag(CRATE_VERSION))
                 .tag(self.base_zkvm_tag("latest"))
-                .bulid_arg("BASE_IMAGE_TAG", self.base_tag(CRATE_VERSION))
-                .exec(&workspace_dir)
+                .bulid_arg("BASE_IMAGE_TAG", self.base_tag(CRATE_VERSION));
```
jsign, line 164: typo: s/bulid_arg/build_arg

crates/ere-openvm/src/lib.rs
```diff
@@ -149,34 +149,63 @@ impl EreOpenVM {
...
-            _resource,
+            resource,
         })
     }
 
-    fn sdk(&self) -> Result<Sdk, CommonError> {
-        let sdk =
-            Sdk::new_without_transpiler(self.app_config.clone()).map_err(CommonError::SdkInit)?;
+    fn cpu_sdk(&self) -> Result<CpuSdk, CommonError> {
+        let sdk = CpuSdk::new_without_transpiler(self.app_config.clone())
```
jsign, line 157: Is there any relevant implication of using `CpuSDK::new_without_transpiler` here compared to the `CpuSDK::new` in the ctor above?

crates/ere-openvm/src/lib.rs
```diff
@@ -149,34 +149,63 @@ impl EreOpenVM {
...
-            .sdk()
-            .map_err(|e| OpenVMError::from(ExecuteError::from(e)))?;
-
         let mut stdin = StdIn::default();
         serialize_inputs(&mut stdin, inputs);
 
         let start = Instant::now();
-        let public_values = sdk
-            .execute(self.app_exe.clone(), stdin)
+        let public_values = with_sdk!(self, |sdk| sdk.execute(self.app_exe.clone(), stdin))
```
jsign, line 208: Do you know if there's any relevant time overhead in the SDK creation?
Mostly thinking if is it worth pulling the SDK creation to the ctor, and avoid that overhead (if any) in execute/prove/etc.

Also, not sure if the SDKs have a common trait -- if that's the case sounds nice since we could avoid this macro if we store the SDK as a field created in ctor.

One benefit is that the way things work now it will only panic if the CUDA feature wasn't provided on execute/prove/etc and not a soon as the ctor is called. But I don't think that's a big deal anyway since low chances it is a relevant detail.

(To be clear, not a strong opinion)

crates/ere-openvm/src/lib.rs
```diff
@@ -228,7 +252,7 @@ impl zkVM for EreOpenVM {
         let proof = VmStarkProof::<SC>::decode(&mut proof)
             .map_err(|e| OpenVMError::from(VerifyError::DeserializeProof(e)))?;
 
-        Sdk::verify_proof(&self.agg_vk, self.app_commit, &proof)
+        CpuSdk::verify_proof(&self.agg_vk, self.app_commit, &proof)
```
jsign, line 255: Interesting, is there a GpuSDK::verify_proof? :smile:

docker/openvm/Dockerfile
```diff
@@ -5,6 +5,26 @@ FROM ${BASE_IMAGE_TAG}
...
+    apt update && \
+    apt install -y cuda-toolkit-12-9 && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*)
+
+# Add nvcc to path
+ENV PATH=/usr/local/cuda/bin:$PATH
+
+# Default to build for RTX 50 series
+ARG CUDA_ARCH=120
+ENV CUDA_ARCH=${CUDA_ARCH}
```
jsign, line 26: Uhm, I was wondering earlier what this `CUDA_ARCH` was for.

Do you know what happens if the host has a 4090 and doesn't set `CUDA_ARCH` to the right value? If the program crashes, then I think that might be okay but might still feel a bit cryptic.

I found https://crates.io/crates/nvml-wrapper/0.11.0 which could be interesting to auto-detect the right thing maybe? Long ago in workload, I did an even [more rawish](https://github.com/eth-act/zkevm-benchmark-workload/blob/0e60e902bed06580766f1f70fc889229ea75bc89/crates/metrics/src/lib.rs#L77) use of the `nvidia-smi`. 

If that could work, might be a more reliable way to avoid misconfigs and forget about this.

jsign (approved):
LGTM! Left some comms for your consideration

crates/ere-dockerized/src/lib.rs
han0110: Fixed in https://github.com/eth-act/ere/pull/113/commits/3340fce0bbf41b70b6dc8805bff9497f15b3f889

jsign:
> > I thin eventually we need to handle the cuda building thing on CI, otherwise I feel the building process will get more and more complicated :(
> 
> By this, you mean requiring a GH runner with a GPU? Or having a more beefy machine to support installing and compiling kernels in a reasonable amount of time? I think if it is the latter, we can ask the devops team to have a more powerful runner than the most powerful documented one. In the docs they say this is possible, but has to be asked explicitly.
> 
> If that could help, I can reach out to the devops team to explore our options.

@han0110, re-ping reg ^ just in case might get lost with my usual verbose pr comments :smile: 
LMK if I can help here.

PR #114, opened 2025-08-29
risc0: Guest program compilation with stock rust compiler.

rodiazet:
This PR implements support for a guest program compilation to the target `riscv32ima-unknown-none-elf` with stock rust compiler. It can be enable by defining environment variable `ERE_GUEST_TOOLCHAIN=nightly` to build using rust nightly toolchain.


crates/ere-risc0/src/compile_stock_rust.rs
```diff
@@ -0,0 +1,123 @@
...
+use cargo_metadata::MetadataCommand;
+use risc0_binfmt::ProgramBinary;
+use std::fs;
+use std::path::Path;
+use std::process::Command;
+use tracing::info;
+
+static CARGO_ENCODED_RUSTFLAGS_SEPARATOR: &str = "\x1f";
+// TODO: Make this with `zkos` package building to avoid binary file storing in repo.
+// File taken from https://github.com/risc0/risc0/blob/v3.0.1/risc0/zkos/v1compat/elfs/v1compat.elf
```
kevaundray, line 12: what is this?

rodiazet: This is "kernel elf" binary which has to be used together with guest program binary elf to make proper risc0 program which can be executed on this zkvm. Using risc0 sdk it was previously done with ` risc0_build::build_package`

kevaundray: Ah right, could we make an issue with the TODO above? And also an issue to have separate Dockerfiles for the stock compilers. These aren't blocking right now or urgent

main thing with the elf file is that if it becomes out of sync

kevaundray (approved):

kevaundray:
LGTM -- waiting for CI to pass

crates/ere-risc0/src/compile_stock_rust.rs
rodiazet: https://linear.app/zkevm/issue/ZKE-294/build-kernel-elf-for-risc0-instead-of-storing-it-in-the-ere-repo
https://linear.app/zkevm/issue/ZKE-295/create-separated-dockerfiles-for-guest-program-build-with-rust-stock

PR #119, opened 2025-09-01
Add clang to ere-base

jsign:
This PR adds `clang` to the installed packages in `ere-base`.

This is required since some dependencies in guest program require clang for building. As a concrete example, this is the case for [Ethrex guest program](https://github.com/eth-act/zkevm-benchmark-workload/pull/158).


han0110 (approved):
LGTM

PR #120, opened 2025-09-02
ere-risc0: alignment test

jsign:


han0110 (approved):
LGTM! Didn't know that only power of 2 alignment is supported in Risc0 (I thought anything >4 should work).

han0110:
I see, that makes sense!

PR #122, opened 2025-09-03
Reuse base image if unchanged in CI

### LINKED ISSUE #2: cache docker image compilations for zkVMs
Compiling each zkVM can vary between 5-10 minutes. This doesn't change with every PR, so we can cache this step until the version is changed.

han0110:
This PR attempts to resolve #2.

In CI we detect if the Dockerfile related files changed or not, if it's not changed, we re-tag the image remotely (so no pulling/pushing). For now the related files are only the Dockerfile itself and the SDK installation script, so I think it's maintainable, but with the related files grow we probably need to find a better way to decide to rebuild or not.

If the `ere-base` and `ere-base-{zkvm}` images are reused, the CI time becomes ~12mins.

### DIFF
--- a/.github/workflows/test-zkvm.yml
+++ b/.github/workflows/test-zkvm.yml
@@ -87,22 +87,64 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Check Dockerfile changes
+        id: changed_files
+        uses: tj-actions/changed-files@v46
+        with:
+          files_yaml: |
+            base:
+              - docker/base/Dockerfile.base
+            base_zkvm:
+              - docker/${{ inputs.zkvm }}/Dockerfile
+              - scripts/sdk_installers/install_${{ inputs.zkvm }}_sdk.sh
+       
       - name: Get image version and tags of ere-base and ere-base-${{ inputs.zkvm }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test-zkvm.yml
```diff
@@ -87,22 +87,67 @@ jobs:
...
+          BUILD_BASE_ZKVM_IMAGE="true"
+
+          # Try to reuse image by re-tagging if related files are unchanged
+          if [ "${{ github.event_name }}" == "pull_request" ]; then
+            PR_BASE_GIT_REV="${{ github.event.pull_request.base.sha }}"
+            PR_BASE_GIT_REV="${PR_BASE_GIT_REV:0:7}"
+            PR_BASE_IMAGE_VERSION="$ZKVM_CRATE_VERSION-$PR_BASE_GIT_REV"
+
+            if [ "${{ steps.changed_files.outputs.base_any_changed }}" == "false" ]; then
+              if docker buildx imagetools create --tag "$BASE_IMAGE_TAG" "$BASE_IMAGE_NAME:$PR_BASE_IMAGE_VERSION" 2>/dev/null; then
```
han0110, line 130: This creates the new tag remotely without pulling/pushing

jsign (approved):

PR #125, opened 2025-09-04
Upgrade ZisK to `0.11.0`

han0110:
- Upgrade ZisK to `0.11.0`
- Now public values are included in proof directly, hence public values deserialization is updated a bit
- Expose environment variables to enables proving flags for fine-tunring, also inherit those envs in `ere-dockerized`


crates/ere-zisk/src/lib.rs
```diff
@@ -642,7 +616,7 @@ mod tests {
     #[test]
     fn test_prove() {
         let program = basic_program();
-        let zkvm = EreZisk::new(program, ProverResourceType::Cpu);
+        let zkvm = EreZisk::new(program, ProverResourceType::Gpu);
```
jsign, line 619: How does this work in CI?

jsign (approved):
LGTM

crates/ere-zisk/src/lib.rs
han0110: Ah I forget to clean it up when testing GPU.

It works in CI because currently runner can't run the `prove` due to insufficient disk space, so this is not caught 🥲

PR #127, opened 2025-09-08
Fix invalid `--chmod` Dockerfile syntax in `docker/sp1`

ilitteri:
The error occurs because the `--chmod=+x` syntax in the Dockerfile is invalid (I'm running Docker Desktop version 4.37.2 in macOS Sequoia 15.6.1). The `--chmod` flag expects an octal string between `0` and `07777`, but `+x` is not a valid octal format.

I opted to use `755` because it worked for me, but we can use `700` if only the owner (root in the container) should execute it.

The same syntax is used in RISC0's and Ziern's Dockerfiles. I didn't updated them because I didn't try them, but if it's ok with you I can fix them too.

The error log I got was the following:

image

### DIFF
--- a/docker/risc0/Dockerfile
+++ b/docker/risc0/Dockerfile
@@ -31,7 +31,7 @@ ENV NVCC_APPEND_FLAGS=${NVCC_APPEND_FLAGS}
 ENV PATH=/usr/local/cuda/bin:$PATH
 
 # Copy and run the Risc0 SDK installer script
-COPY --chmod=+x scripts/sdk_installers/install_risc0_sdk.sh /tmp/install_risc0_sdk.sh
+COPY --chmod=755 scripts/sdk_installers/install_risc0_sdk.sh /tmp/install_risc0_sdk.sh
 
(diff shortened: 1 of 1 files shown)


han0110 (commented):
Thanks for the fix, didn't know this is a new feature of docker buildx. I think it makes sense to use `--chmod=755` for backward compatibility, it'd be nice if you could also update other dockerfiles as well.

ilitteri:
> Thanks for the fix, didn't know this is a new feature of docker buildx. I think it makes sense to use `--chmod=755` for backward compatibility, it'd be nice if you could also update other dockerfiles as well.

You're welcome! The other fixes were done in [04af88f](https://github.com/eth-act/ere/pull/127/commits/04af88f20bab3c148fcbe03b4ae33f23383bf88d) and [e342e80](https://github.com/eth-act/ere/pull/127/commits/e342e800ba9aa8d82ad22ef8a9a4531c90c70fbf).

han0110 (approved):
LGTM

PR #136, opened 2025-09-15
feat: `ere-miden` 

brech1:
# Description

I would address this PR as an initial `miden-vm` integration. One of the biggest differentiators of it with other `ere` crates is that guests are implemented in `miden assembly`. The reason for this is that compiler support for [Rust to Masm is being rolled out](https://miden.xyz/resource/blog/compiler-release-04) but not fully compatible at the moment. 

## Then why Miden?

Adding a bit of context, the [client side proving benchmarks](https://github.com/privacy-ethereum/csp-benchmarks/tree/CSP-Q3-2025/zkvm) now have `ere` as its default zkVM interface. This allows us to support all `ere` zkVMs with low effort and a standardize way of interacting with them.

The `miden-vm` has client-side proving capabilities, with comparable [results](https://github.com/privacy-ethereum/csp-benchmarks/pull/32) to other zkVMs, so we should have it as a compatible host on our end.

If the goal of `ere` is to provide an interface for benchmarking zkVMs then this is a good addition, but it's good to point out that Miden is not ready to be used to prove Ethereum's STF using Rust guests.

## IO

`miden-vm` public inputs are a `[u64;16]` element, and called a `stack` (since that's what it is internally). Private inputs are placed on an `advice tape`, a dynamic size vector.

For the parsing of the inputs I'm deserializing all things and putting them in the `advice`. So, not dealing with public inputs at the moment.

Miden-vm output is what's left in the `stack` after execution. Verification method only requires public inputs, serialized inside the proof bundle.

## Dockerized Miden

Is not implemented. The `miden-core` crate hosts an executable CLI with capabilities similar to other zkVM SDKs, but the standard setup doesn't need it to run since the crate is being used directly.

Implementing compilation through this CLI could however enable more configurations and provide a more standard `masm` project setup, like having the inputs set in a file and set stack and advice inputs directly.

I would say the main reason not to take this route is that ideally, the `Rust -> Miden Assembly` route should replace the current setup, with my main argument being that all other zkVMs are being benchmarked through a higher level language like Rust and don't count with specific optimizations (unless of course specific precompiles are implemented) and secondly that it's less likely to expect application developers to adopt an assembly-like language to build their apps rather than Rust.

## Tests

I wasn't able to implement the standard host `test-utils`. Since the output of the `miden-vm` is a `16` element stack of u16, I don't think it's expected to be handling objects in the stack, so `deserialization` utilities didn't make much sense here, other than regrouping the field elements from the `u8`s. However, this could change if guests are implemented in Rust.

## Contribution Guidelines

I tried to follow `ere`'s general structure and practices but anything is up to change.

### DIFF
--- a/.github/workflows/test-zkvm-miden.yml
+++ b/.github/workflows/test-zkvm-miden.yml
@@ -1,0 +1,20 @@
+name: Test and clippy Miden
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  test:
+    uses: ./.github/workflows/test-miden.yml
+    permissions:
+      contents: read
+      packages: write
+    with:
+      zkvm: miden
+      toolchain: 1.88.0
+      test_ere_dockerized: false
+      default_features: false
+      test_options: ""
--- a/.github/workflows/test-zkvm-miden.yml
+++ b/.github/workflows/test-zkvm-miden.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   test:
-    uses: ./.github/workflows/test-miden.yml
+    uses: ./.github/workflows/test-zkvm.yml
     permissions:
       contents: read
       packages: write
--- a/.github/workflows/test-zkvm-miden.yml
+++ b/.github/workflows/test-zkvm-miden.yml
@@ -16,5 +16,5 @@ jobs:
       zkvm: miden
       toolchain: 1.88.0
       test_ere_dockerized: false
-      default_features: false
+      default_features: true
       test_options: ""
--- a/crates/ere-miden/src/compile.rs
+++ b/crates/ere-miden/src/compile.rs
@@ -1,0 +1,67 @@
+use crate::{
+    MidenProgram,
+    error::{CompileError, MidenError},
+};
+use miden_assembly::Assembler;
+use miden_core::utils::Serializable;
+use miden_stdlib::StdLibrary;
+use std::{fs, path::Path};
+use zkvm_interface::Compiler;
+
+#[allow(non_camel_case_types)]
+pub struct MIDEN_TARGET;
+
+impl Compiler for MIDEN_TARGET {
+    type Error = MidenError;
+    type Program = MidenProgram;
+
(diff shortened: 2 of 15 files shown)


.github/workflows/test-zkvm-miden.yml
```diff
@@ -0,0 +1,20 @@
...
+  test:
+    uses: ./.github/workflows/test-miden.yml
+    permissions:
+      contents: read
+      packages: write
+    with:
+      zkvm: miden
+      toolchain: 1.88.0
+      test_ere_dockerized: false
+      default_features: false
```
han0110, line 19: We could set it to `true` becasue there is nothing block the CI to run with the default feature
```suggestion
      default_features: true
```

crates/ere-miden/src/compile.rs
```diff
@@ -0,0 +1,64 @@
...
+    type Error = MidenError;
+    type Program = MidenProgram;
+
+    fn compile(&self, guest_directory: &Path) -> Result<Self::Program, Self::Error> {
+        let dir_name = guest_directory
+            .file_name()
+            .and_then(|name| name.to_str())
+            .ok_or(CompileError::InvalidProgramPath)?;
+
+        let entrypoint = format!("{}.masm", dir_name);
```
han0110, line 21: To make clippy happy
```suggestion
        let entrypoint = format!("{dir_name}.masm");
```

docker/miden/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ENV MIDEN_VERSION="v0.17.1" \
+    MIDEN_TOOLCHAIN_VERSION="1.88.0"
+
+# Miden CLI Installation
+# COPY --chmod=755 scripts/sdk_installers/install_miden_sdk.sh /tmp/
+# RUN /tmp/install_miden_sdk.sh && rm /tmp/install_miden_sdk.sh
+
+# Verify
+# RUN miden-vm --version
+
```
han0110, line 15: To override the default 1.86.0 which is too old for miden
```suggestion
RUN rustup default 1.88.0
```

han0110 (approved):
LGTM! Thanks again for the nice PR!

PR #137, opened 2025-09-16
Use fork of `zisk` that allows to override `CUDA_ARCH`

### LINKED ISSUE #134: zisk: detection of GPU arch fails for L40s
I was trying ZisK in our L40 machine, but I got the following runtime error:

Image

Digging into our generated `ere-base-zisk` image, I noticed:

```
root@e2f53b8474d0:/app/zisk# cuobjdump --list-elf ~/.zisk/target/release/bin/libzisk_witness_gpu.so
ELF file    1: tmpxft_000032b4_00000000-0.sm_90.cubin
ELF file    2: tmpxft_000032b4_00000000-1.sm_90.cubin
ELF file    3: tmpxft_000032b4_00000000-2.sm_90.cubin
ELF file    4: tmpxft_000032b4_00000000-3.sm_90.cubin
ELF file    5: tmpxft_000032b4_00000000-4.sm_90.cubin
ELF file    6: tmpxft_000032b4_00000000-5.sm_90.cubin
ELF file    7: tmpxft_000032b4_00000000-6.sm_90.cubin
root@e2f53b8474d0:/app/zisk#
```

Which looks pretty bad since:

```
$ nvidia-smi --query-gpu=name,compute_cap --format=csv,noheader
NVIDIA L40S, 8.9
```

So the `libzisk-witness.so` is compiled for SM90 but L40 is SM89.

Navigating into the `zisk` repo, this library is actually built in another dependency [pil2-proofman](https://github.com/0xPolygonHermez/pil2-proofman). There inside `pil2-stark/Makefile` there’s some logic that ultimately calls a [src/goldilocks/configure.sh](https://github.com/0xPolygonHermez/pil2-proofman/blob/a0ec36d1e5eb58ce59fe5805bc5856092cfbd600/pil2-stark/Makefile#L4) which does the GPU arch detection.

I run this manually and was failing since it didn’t find `nvcc` in the PATH, thus it ends up using the default arch set in [CudaArch.mk which is SM90](https://github.com/0xPolygonHermez/pil2-proofman/blob/a0ec36d1e5eb58ce59fe5805bc5856092cfbd600/pil2-stark/src/goldilocks/CudaArch.mk#L1) (a bit unfortunate there's a default value, since it would be better if this is empty and the whole thing crashes).

`nvcc` is actually installed in the container but it is missing in the PATH — if we fix this, that part would be solved.

But unfortunately, that's not the end of the story... `docker build` apparently doesn't allow access to the underlying GPU (i.e., there's no equivalent `--gpus all` that you can use in `docker run`). This means that the call to `nvcc` will fail and will always default to the default SM90.

Said differently, _any_ image build that relies on accessing the GPU should probably misbehave. I guess for most zkVMs that was fine since they didn't have auto-detection of the arch and they're passed as env (e.g., [as we do with OpenVM](https://github.com/eth-act/ere/blob/2f979dbd01fc11a08a6c9d650f5f8fad698e8469/docker/openvm/Dockerfile#L25))

I think if ZisK can support setting arch via envs, it would be ideal and quite elegant, like OpenVM.

---

To avoid this blocking me from my intentions to run ZisK in L40s, I forked ZisK and `pil2-proofman`, such that ZisK uses my `pil2-proofman` and my `pil2-proofman` has a changed default `CudaArch.mk` -- such that when the image is built and the `nvcc` call fails, it will fallback to the SM89 that I want for L40.

I can confirm that now the shared screenshot crash doesn't happen and the GPU is reporting working proving stuff.

han0110:
Attempt to resolve #134 

This PR:

- Use fork of `zisk` to build `libzisk_witness_gpu.so` that allows to override `CUDA_ARCH` (see the diff [here](https://github.com/han0110/pil2-proofman/commit/89f018a85cbe3b9cb9bac8a972fa19de84dbb80d))
- Inherit the env `CUDA_ARCH` in `ere-dockerized` in the same way to do for OpenVM

### DIFF
--- a/crates/ere-dockerized/src/cuda.rs
+++ b/crates/ere-dockerized/src/cuda.rs
@@ -1,5 +1,5 @@
 use std::{env, process::Command};
-use tracing::info;
+use tracing::{info, warn};
 
 /// Returns Cuda GPU compute capability, for example
 /// - RTX 50 series - returns `12.0`
@@ -33,12 +33,16 @@ pub fn cuda_compute_cap() -> Option<String> {
 ///
 /// Otherwise it returns `None`.
 pub fn cuda_arch() -> Option<String> {
-    if let Ok(cuda_arch) = env::var("CUDA_ARCH")
-        && cuda_arch.starts_with("sm_")
-        && cuda_arch[3..].parse::<usize>().is_ok()
-    {
-        info!("Using CUDA_ARCH {} from env variable", cuda_arch);
-        Some(cuda_arch)
+    if let Ok(cuda_arch) = env::var("CUDA_ARCH") {
+        if cuda_arch.starts_with("sm_") && cuda_arch[3..].parse::<usize>().is_ok() {
+            info!("Using CUDA_ARCH {cuda_arch} from env variable");
+            Some(cuda_arch)
+        } else {
+            warn!(
+                "Skipping CUDA_ARCH {cuda_arch} from env variable (expected to be in format `sm_XX`)"
+            );
+            None
+        }
     } else if let Some(cap) = cuda_compute_cap() {
         info!("Using CUDA compute capability {} detected", cap);
         Some(format!("sm_{}", cap.replace(".", "")))


jsign (commented):
@han0110, a priori it makes sense to me (changes here and pil2-stark forks), but I did a quick test and I'm not sure it is running correctly:
- Cloned workload and changed ere commit to `6d332745383e16d61316f85221f020a0ce7b9a11`
- `export CUDA_ARCH=89` 
- Run the proving

And I get the original error in first post in https://github.com/eth-act/ere/issues/134.

image


No docker images were cached since its a new ere commit and saw the whole thing built from scratch.

Tried twice to double check and seems to be the case -- did you happen to confirm it in some 4090 or L40s machine?  

It could be the case that the `nvcc` call to access the GPU happens before that CUDA_ARCH usage, so the `.mk` file doesn't get edited as expected.

jsign:
@han0110 got this now:
image

jsign (approved):
Tested again and looks good!

han0110:
Thanks for the help of testing on L40 🙏

PR #139, opened 2025-09-17
Complete `ere-pico`

han0110:
This PR:

- Copy and modify the client from https://github.com/brevis-network/pico/blob/v1.1.7/sdk/sdk/src/client.rs and remove the groth16 wrapper
- Add `PicoProofWithPublicValues` because the public values of recursive proof only contains digest of guest public values.
- Check public values digest matches in method `verify`, and returns public values from `PicoProofWithPublicValues`

This should unblock https://github.com/eth-act/zkevm-benchmark-workload/issues/173

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,8 +78,8 @@ openvm-stark-sdk = { git = "https://github.com/openvm-org/stark-backend.git", ta
 openvm-transpiler = { git = "https://github.com/openvm-org/openvm.git", tag = "v1.4.0" }
 
 # Pico dependencies
+pico-p3-field = { git = "https://github.com/brevis-network/Plonky3.git", package = "p3-field", rev = "a4d376b" }
 pico-vm = { git = "https://github.com/brevis-network/pico.git", tag = "v1.1.7" }
-pico-sdk = { git = "https://github.com/brevis-network/pico.git", tag = "v1.1.7" }
 
 # Risc0 dependencies
 risc0-build = "3.0.3"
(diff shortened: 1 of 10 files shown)


crates/ere-pico/src/client.rs
```diff
@@ -0,0 +1,99 @@
+// Copied and modified from https://github.com/brevis-network/pico/blob/v1.1.7/sdk/sdk/src/client.rs.
+// The `EmbedProver` is removed because we don't need the proof to be verified
+// on chain.
```
jsign, line 3: Worth creating an issue to remember trying to avoid/remove this eventually?

I think might be useful to ping Brevis to provide proper functionality via the SDK.
(I added you to a TG group `Brevis <> EF` which can be a good medium to ask this or anything, if you want)

Cargo.toml
```diff
@@ -78,8 +78,8 @@ openvm-stark-sdk = { git = "https://github.com/openvm-org/stark-backend.git", ta
 openvm-transpiler = { git = "https://github.com/openvm-org/openvm.git", tag = "v1.4.0" }
 
 # Pico dependencies
+pico-p3-field = { git = "https://github.com/brevis-network/Plonky3.git", package = "p3-field", rev = "a4d376b" }
 pico-vm = { git = "https://github.com/brevis-network/pico.git", tag = "v1.1.7" }
-pico-sdk = { git = "https://github.com/brevis-network/pico.git", tag = "v1.1.7" }
```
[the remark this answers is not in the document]
han0110: yeah, if pico is willing to add another function `prove_compressed` and uncomment the verification checks, we can add back the sdk and use it for simplifity.

crates/ere-pico/src/client.rs
han0110: Make sense, created an issue in https://github.com/eth-act/ere/issues/140, and will ask if the Brevis team can help us on that

PR #141, opened 2025-09-18
Test dockerized pico

han0110:
Add the missed dockerized pico test that should be added in https://github.com/eth-act/ere/pull/139


han0110:
Merging without approval because it's only adding tests

PR #37, opened 2025-06-20
[draft] Cg/add airbender

codygunton:

PR #175, opened 2025-10-20
Integrate Airbender

han0110:
Integrates https://github.com/matter-labs/zksync-airbender/tree/v0.5.0 into crate `ere-airbender`

Supersedes #37

### DIFF
--- a/crates/zkvm/airbender/src/client.rs
+++ b/crates/zkvm/airbender/src/client.rs
@@ -69,7 +69,7 @@ impl AirbenderSdk {
             });
         }
 
-        // Parse public values from stdout in format of:
+        // Parse public values 8 u32 words (32 bytes) from stdout in format of:
         // `Result: {v0}, {v1}, {v2}, {v3}, {v4}, {v5}, {v6}, {v7}`
         let public_values = output
             .stdout


.github/workflows/test-zkvm-airbender.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  test:
+    uses: ./.github/workflows/test-zkvm.yml
+    permissions:
+      contents: read
+      packages: write
+    with:
+      zkvm: airbender
+      toolchain: nightly
+      skip_prove_test: true
```
han0110: It takes 33.21 GB to run the prove test, so skipped here.

crates/zkvm/airbender/src/client.rs
```diff
@@ -0,0 +1,244 @@
...
+fn extract_public_values_and_vk_hash_chain(
+    proof: &ProgramProof,
+) -> Result<(PublicValues, VkHashChain), AirbenderError> {
+    if proof.register_final_values.len() != 32 {
+        return Err(AirbenderError::InvalidRegisterCount(
+            proof.register_final_values.len(),
+        ));
+    }
+
+    let public_values = proof.register_final_values[10..18]
```
jsign: Interesting -- I see.

tests/airbender/basic/src/airbender_rt.rs
```diff
@@ -0,0 +1,60 @@
+use core::alloc::{GlobalAlloc, Layout};
```
jsign: Huh, I see -- I wonder why runtime stuff this isn't part of some crate in their SDK?

crates/zkvm/airbender/src/compiler/rust_rv32ima.rs
```diff
@@ -0,0 +1,97 @@
...
+            .toolchain(toolchain)
+            .build_options(CARGO_BUILD_OPTIONS)
+            .rustflags(RUSTFLAGS)
+            .exec(guest_directory, TARGET_TRIPLE)?;
+        let bin = objcopy_binary(&elf)?;
+        Ok(bin)
+    }
+}
+
+fn objcopy_binary(elf: &[u8]) -> Result<Vec<u8>, AirbenderError> {
```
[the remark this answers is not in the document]
han0110: I'm not sure the benefit to take binary instead of elf here tbh (perhaps they just don't need the section metadata).

tests/airbender/basic/src/airbender_rt.rs
han0110: I think their guest SDK is very minimum, they don't even provide a default working allocator (their default is a panicing one), so here we have enable the feature `custom_allocator`, then implement our own one.

Their SDK mainly provides read input words, and commit output 32 bytes.

PR #181, opened 2025-10-28
Update Jolt to `v0.3.0-alpha`

### LINKED ISSUE #90: Update Jolt to v0.3.0-alpha
https://github.com/a16z/jolt/tree/v0.3.0-alpha

han0110:
- Upgrade Jolt to `v0.3.0-alpha`
- Compiler target is updated from `rv32ima` to `rv64imac`
- Refactored the `jolt_methods.rs` into `sdk.rs`
- Accept the fact that Jolt hasn't support private input, now we include the input in the proof so the end-to-end flow works, we could remove them when the private input is supported.

Resolves #90

### DIFF
--- a/.github/workflows/test-zkvm-zisk.yml
+++ b/.github/workflows/test-zkvm-zisk.yml
@@ -15,5 +15,4 @@ jobs:
     with:
       zkvm: zisk
       toolchain: 1.86.0
-      free_up_disk_space: true
       skip_prove_test: true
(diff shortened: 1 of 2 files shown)


crates/zkvm/airbender/src/lib.rs
```diff
@@ -146,7 +146,6 @@ mod tests {
         let zkvm = EreAirbender::new(program, ProverResourceType::Cpu).unwrap();
 
         let test_case = BasicProgramInput::valid().into_output_sha256();
-        run_zkvm_execute(&zkvm, &test_case);
```
han0110: Removed a copy paste leftover

PR #182, opened 2025-10-28
Add 'get_verifying_key' API to ZkVM

wraitii:
This add `fn get_verifying_key(&self) -> anyhow::Result<Self::VerifyingKey>` to the ZkVM trait, which is the necessary counterpart of proofs for verification.

I return an associated trait that supports `Serialize` for generic use, can potentially add more bounds. I currently clone the data-types under the assumption they are cheap to clone & returning references is annoying.

Went with `get_verifying_key` as that's generally the convention for this as far as I know, but Risc0 uses "image ID" & Cairo used "class hash" but that was kinda starknet-specific.

Alternatives:
- Could be added to the compiler or to a new "GuestProgram" trait, but many ZkVMs need some initialisation to return the verifying key so putting it in the zkvm seems natural.
- Could just expose a way to serde the verifying key.
- Could not be part of the trait, just a common helper.

Issues with my PR:
- Haven't found if the concept exists for jolt & nexus.
- Miden doesn't implement `Serialize` so I return `Vec<u8>`

### DIFF
--- a/crates/zkvm/openvm/src/lib.rs
+++ b/crates/zkvm/openvm/src/lib.rs
@@ -272,6 +272,14 @@ mod tests {
     }
 
     #[test]
+    fn test_get_verifying_key() {
+        let program = basic_program();
+        let zkvm = EreOpenVM::new(program, ProverResourceType::Cpu).unwrap();
+
+        run_zkvm_get_verifying_key(&zkvm);
+    }
+
+    #[test]
     fn test_prove_invalid_input() {
         let program = basic_program();
         let zkvm = EreOpenVM::new(program, ProverResourceType::Cpu).unwrap();
--- a/crates/dockerized/dockerized/src/lib.rs
+++ b/crates/dockerized/dockerized/src/lib.rs
@@ -646,6 +646,19 @@ mod test {
         };
     }
 
+    macro_rules! test_get_verifying_key {
+        ($zkvm:ident) => {
+            #[test]
+            fn test_get_verifying_key() {
+                let (_guard, zkvm) = zkvm();
+
+                run_zkvm_get_verifying_key(&zkvm);
+
+                drop(zkvm);
+            }
+        };
+    }
+
     mod airbender {
         test_compile!(Airbender, "basic");
         test_execute!(
@@ -658,6 +671,7 @@ mod test {
             BasicProgramInput::valid().into_output_sha256(),
             [Vec::new(), BasicProgramInput::invalid().serialized_input()]
         );
+        test_get_verifying_key!(Airbender);
     }
 
     mod jolt {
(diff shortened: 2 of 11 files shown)


han0110 (commented):
LGMT. 

Although I think the conept of verifying key here is hard to define, because some data required during verificartion is read from pre-installed files, or baked into rust code, the verifying key implemented in this PR is more like some kind of "program dependant setup" to me (but we could rename it later if we find a better name).

To fix the CI error, we could leave `EreDockerizedzkVM`'s `get_verifying_key` as `unimplemented` and we could do that later.

crates/zkvm/openvm/src/lib.rs
```diff
@@ -188,6 +188,11 @@ impl zkVM for EreOpenVM {
         Ok(public_values)
     }
 
+    type VerifyingKey = AggVerifyingKey;
+    fn get_verifying_key(&self) -> anyhow::Result<Self::VerifyingKey> {
+        Ok(self.agg_vk.clone())
```
han0110: For OpenVM I think we should also include the `app_commit` as the `VerifyingKey`, because the `agg_vk` is universal, only the `app_commit` is related/generated from the program.

wraitii:
> Although I think the conept of verifying key here is hard to define, because some data required during verificartion is read from pre-installed files, or baked into rust code, the verifying key implemented in this PR is more like some kind of "program dependant setup" to me (but we could rename it later if we find a better name).

For what it's worth on Hyli we're calling this the "program ID" seems it is indeed intended to be the identifier of the program being run on the VM... I do agree this is kind of a confusing thing.

> To fix the CI error, we could leave `EreDockerizedzkVM`'s `get_verifying_key` as `unimplemented` and we could do that later.

Let me check if I can fix that, I wasn't sure how to test everything :)

crates/zkvm/openvm/src/lib.rs
han0110: Yeah `app_commit` seems closer to what we want, we should remove `agg_vkey` here.

>  It sounds like it can be reconstructed by anyone

Yes

wraitii:
Updated the PR:
- Added an endpoint for querying the verifying key via docker, returning vec<u8>
  - And call it in the proving test, hopefully that works
- Added a `MidenProgramInfo` serde wrapper + removed comment
- Switch to app_commit, indeed agg_vkey is just reconstructible.

crates/dockerized/server/proto/api.proto
```diff
@@ -62,3 +63,17 @@ message VerifyResponse {
 message VerifyOk {
     bytes public_values = 1;
 }
+
+message GetVerifyingKeyRequest {
```
han0110: The server also needs to be update to implement this endpoint

PR #206, opened 2025-11-05
TamaGo compilation support

marcinbugaj:
This PR:
- adds an installation script for TamaGo
- adds `impl Compiler` for the TamaGo compiler for Zisk
- provides a tests
- modifies the basic image to include TamaGo compiler

### DIFF
--- a/crates/dockerized/dockerized/src/lib.rs
+++ b/crates/dockerized/dockerized/src/lib.rs
@@ -753,7 +753,7 @@ mod test {
     }
 
     mod zisk {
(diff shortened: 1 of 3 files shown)


han0110 (commented):
LGTM!

To fix the CI error, we'd need to fix [this line](https://github.com/eth-act/ere/blob/master/crates/dockerized/dockerized/src/lib.rs#L756) to `basic_rust`, and add the tamago `go` to the path in dockerfile I think

Issue #207, opened 2025-11-05
OpenVM: GPU proving seem to be failing


jsign:
When I try to prove a mainnet block using workload, I get the following output:
```
{
  "name": "rpc_block_23326233",
  "timestamp_completed": "2025-11-05T19:26:23.100858959Z",
  "metadata": {
    "block_used_gas": 11228935
  },
  "proving": {
    "crashed": {
      "reason": "zkVM method error: OpenVM proving failed: VM error: execution error: execution failed at pc 12, err: DebugPanic"
    }
  }
```

Notes:
- It runs for quite a long time, but seems to fail, I suspect at the end of proving.
- [Logs of the run](https://gist.githubusercontent.com/jsign/ba2df392ae4225db6b69d9f735659ed6/raw/454d42d78e9d1a5dfd569e47e8aed32f66823806/openvm.logs).
- Maybe obvious: I didn't pull any pre-generated Docker. For GPU proving, I always let them be generated in the machine for full efficiency.
- It failed for 30 mainnet blocks (all I tried), so it isn't related to a particular one.
- Since this is `workload@master`, it means that it passed execution-only for mainnet blocks, which sounds quite interesting since might be only a proving problem?

Steps to reproduce:
- Pull latest workload `master`. (Current [last commit](https://github.com/eth-act/zkevm-benchmark-workload/commit/313ef8656e54c7761e8db366968ed7eb00c3de01))
- Download [this mainnet block input file](https://drive.google.com/file/d/15WQCzIW0Ny7K_tXuar7Ga7-doI-MtOpC/view?usp=sharing), and put it inside a folder named `zkevm-fixtures-input`. (Recall this mainnet block isn't particular or similar, fails with any apparently)
- Do `export OPENVM_RUST_TOOLCHAIN="nightly-2025-08-07"`
- Run the proving: `RUST_LOG=info RUSTFLAGS="-C target-cpu=native -C target-feature=+avx512f" cargo run --release -p ere-hosts -- --zkvms openvm --action prove --resource gpu stateless-validator --execution-client reth` (avx512f if supported, but just sharing exactly my command)

When the proving finishes, you should see a single json file in the `zkevm-metrics/reth/openvm` folder with the same content of the first code snippet I shared above.

Note: this is not using the PR of OpenVM with the new precompiles. I also tested in that branch and fails with the same reason, but because it also fails in `master` (i.e., without the new precompiles) sounds it is unrelated)

jsign:
[...]

PR #208, opened 2025-11-06
Remove default value of `RUSTFLAGS`

han0110:
Currently we have `RUSTFLAGS="-C target-cpu=native"` by default to build the `ere-server` in `Dockerfile.server`, but the `target-cpu=native` seems not always work even we build and run the image on the same machine, which leads the #207.

This PR remove the default value and allow to set it via env variable when building the image.

Also suggested by Jonathan, when building in docker we should set the `target-cpu` explicitly of the machine (e.g. `RUSTFLAGS="-C target-cpu=znver3"`).

### DIFF
--- a/docker/risc0/Dockerfile.base
+++ b/docker/risc0/Dockerfile.base
@@ -9,6 +9,7 @@ FROM $BASE_CUDA_IMAGE AS base_cuda
 FROM base${CUDA:+_cuda}
 
 ARG CUDA
+ARG RUSTFLAGS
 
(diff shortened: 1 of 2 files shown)


jsign:
Gave this a try using https://github.com/eth-act/zkevm-benchmark-workload/pull/220

Ran with `RUST_LOG=info RUSTFLAGS="-C target-cpu=znver5 -C target-feature=+avx512f" cargo run --release -p ere-hosts -- --zkvms openvm --action prove --resource gpu stateless-validator --input-folder test-mainnet-blocks/ --execution-client reth` building the images from scratch, but seems like there is still the same problem.

Might worth double-checking, but just sharing the quick test results.

(Maybe this would have solved the "second bullet" problem about proof verification though)

PR #214, opened 2025-11-19
Fix ZisK server recovering

### LINKED ISSUE #212: zisk: server get's permanently down after some proving failure
It looks like that when proving multiple blocks, if some of them fails/panics then the `server` gets down and no further block being proven after (using the same server) work.

To reproduce:
1. Clone latest `workload` repo
2. `RAYON_NUM_THREADS=10 RUST_LOG=info cargo run -p witness-generator-cli --release -- tests --include 1M --include Prague --tag benchmark@v0.0.6` to generate some set of blocks to prove, so some of this will make the zisk emulator fail (prob because of overflowing max cycles)
3. Try proving this set with `cargo run --release -p ere-hosts -- --zkvms zisk --action prove --resource gpu stateless-validator --execution-client reth`

So it starts with proving without problems until some errors like this starts appearing:
```
$ cat zkevm-metrics/reth/zisk-v0.13.0/* | grep crashed -A 3
    "crashed": {
      "reason": "zkVM method error: Command `\"cargo-zisk\" \"prove-client\" \"status\"` exit with exit status: 1\nstderr: Error: Error executing ProveClient command\n\nCaused by:\n    Faile
d to parse server response: EOF while parsing a value at line 1 column 0\n    Raw: \n"
    }
  }
--
    "crashed": {
      "reason": "zkVM method error: Command `\"cargo-zisk\" \"prove-client\" \"prove\" \"--input\" \"/tmp/.tmpoGqUdR/input\" \"--output_dir\" \"/tmp/.tmpoGqUdR/output\" \"-p\" \"ere\" \"--ag
gregation\" \"--verify_proofs\"` exit with exit status: 1\nstderr: Error: Error executing ProveClient command\n\nCaused by:\n    Failed to connect to server: Connection refused (os error 111
)\n"
    }
  }

```
So the first crash error is expected since was a modexp case that probably got a crash error from the zisk emulator since it overflows max cycles in Zisk, so maybe that's okay. But the following error connection refused, seems to be for a fixture that shouldn't crash.

han0110:
Previously we recover the server by recreating, but because the `drop` is not explictly done before recreating, the `cargo-zisk prove-client shutdown` command is done after we create the new server, which makes the following `prove` fail to connect to the server.

Resolves #212

### DIFF
--- a/crates/zkvm/zisk/src/zkvm.rs
+++ b/crates/zkvm/zisk/src/zkvm.rs
@@ -47,6 +47,8 @@ impl EreZisk {
             // Recreate the server if it has been created but failed to get status.
             Some(s) => {
                 if s.status().is_err() {
+                    // Drop server first to make sure the server is shutdown.
+                    drop(server.take());
                     *server = Some(self.sdk.server()?);
                 }
             }


jsign:
I can update workload temporarily to this branch and run the hole thing I was running to double check.

jsign:
I'll wait for the whole thing to finish, re-run these one that timed out and we can see if this is repeatable for some reason or just flakyness. I'll report back later.

han0110:
Could you share after which test crash it becomes like this? I'm guess it's some process not cleaned up properly so the server fails to start

PR #218, opened 2025-11-20
ZisK server cleanup

han0110:
- Cleanup asm service and shared memory manually if ZisK server crashes (this resolves most of the timeout issue)
- Recreate and server if it's not running, with max retry 3 times (in few cases retrying is necessary, not sure why...)

Ran the 10M suites with > 400 tests and there is no longer timeout

### DIFF
--- a/crates/zkvm/zisk/src/zkvm/sdk.rs
+++ b/crates/zkvm/zisk/src/zkvm/sdk.rs
@@ -538,7 +538,9 @@ fn rom_setup(elf_path: &Path) -> Result<RomDigest, Error> {
     Ok(rom_digest)
 }
 
+/// Send shutdown request to ZisK asm services.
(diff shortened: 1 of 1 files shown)


crates/zkvm/zisk/src/zkvm/sdk.rs
```diff
@@ -518,6 +538,38 @@ fn rom_setup(elf_path: &Path) -> Result<RomDigest, Error> {
     Ok(rom_digest)
 }
 
+/// Send shutdown request to ZisK asm services.
+fn shutdown_asm_service(port: u16) {
```
han0110: When ZisK server crashes, the asm services are not cleanup properly, here we do it manually and it seems to resolves most of the server initialization halting issue

jsign (approved):
LGTM. Great! I'll give it a try.

PR #224, opened 2025-11-25
Add `print` and `cycle_scope` methods for `Platform`

han0110:
This PR extends the `Platform` for `zkevm-benchmark-workload` guest usage.

### DIFF
--- a/crates/zkvm/sp1/platform/src/lib.rs
+++ b/crates/zkvm/sp1/platform/src/lib.rs
@@ -2,7 +2,7 @@
 
 extern crate alloc;
 
-use alloc::vec::Vec;
+use alloc::{format, vec::Vec};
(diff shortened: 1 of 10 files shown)


crates/zkvm/sp1/platform/src/lib.rs
```diff
@@ -21,6 +21,19 @@ impl<H: OutputHasher> Platform for SP1Platform<H> {
 
     fn write_whole_output(output: &[u8]) {
         let hash = H::output_hash(output);
-        sp1_zkvm::io::commit_slice(&*hash);
+        sp1_zkvm::io::commit_slice(&hash);
+    }
+
+    fn print(message: &str) {
```
jsign: Nice, this looks better than the log fmting trick we have in workload.

Issue #222, opened 2025-11-24
SP1: unstable runs with 10M EEST tests


jsign:
Looks like SP1 has some instability in server readiness.

For many cases, which are not modexp or other heavy ones, I get the following result:
```
{
  "name": "test_arithmetic.py::test_arithmetic[fork_Prague-benchmark-gas-value_10M-blockchain_test-opcode_ADD-]",
  "timestamp_completed": "2025-11-24T16:13:36.897070227Z",
  "metadata": {
    "block_used_gas": 10000000
  },
  "proving": {
    "crashed": {
      "reason": "RPC to zkVM server error: error InvalidArgument: error sending request for url (http://127.0.0.1:4181/twirp/api.ZkvmService/Prove) (rust_error: \"reqwest::Error { kind: Request, url: \\\"http://127.0.0.1:4181/twirp/api.ZkvmService/Prove\\\", source: hyper_util::client::legacy::Error(Connect, ConnectError(\\\"tcp connect error\\\", 127.0.0.1:4181, Os { code: 111, kind: ConnectionRefused, message: \\\"Connection refused\\\" })) }\")"
    }
  }
}
```

I got this for [473 cases of ~600](https://eth-act.github.io/zkevm-benchmark-runs/?mode=proving&hardware=1xL40s&config=10M-gas-limit&elClient=reth).

(If useful, the L40 machine is free)

jsign:
```
jsign@block-proving-bench-l40s:~/zkevm-benchmark-workload-eest$ cat zkevm-metrics/reth/sp1-v5.2.2/* | grep crashed | wc -l
473
```
Looks like after one case pushes things too much, the rest follow crashes. :(

han0110:
I see, I can also see the leftover `moongate-server` which means the `ere-server` panicked and didn't cleanup that. Will make a PR to try to fix this soon.

PR #225, opened 2025-11-25
Retry RPC and restart server container when necessary

han0110:
Attempt to resolvve #222

- Retry RPC if the error is related to connection error
- When it's a connection error, check if container exists or not, if not we recreate one

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2283,7 +2283,7 @@ dependencies = [
  "iana-time-zone",
  "num-traits",
  "serde",
- "windows-link",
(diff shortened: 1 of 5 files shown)


jsign:
Oh okay, just in time since I was doing the final check before moving away from keys :)
I'll update to 1538ac547b70d03b800c523a9de37c864b00159e and continue the run for the night.

han0110:
> Has been running without crashing all the night, so I think the fix is working. Can report back eventually if this isn't the case, but sounds safe to merge.

Thanks 🙏

Issue #234, opened 2025-12-03
Misleading documents about how input is passed to zkVM host


han0110:
The [README says](https://github.com/eth-act/ere?tab=readme-ov-file#input-handling):

> The input is opaque to `zkVM` and will be passed as is, de/serialization needed to be handled by guest/host themselves. 

But in fact for Airbender and Risc0, the `input` passed to `zkVM::execute` and `zkVM::prove` is prepended with a length before sending to host, for the former it's necessary for the `AirbenderPlatform` to know how many words to read out (there is no way to get the length of the input), for latter it's for `Risc0Platform` to read the whole input without needing to reallocating and copying (no way to get the length of the input as well).

PR #235, opened 2025-12-05
Refactor docs and make `Input` API more explicit

han0110:
- Make `Input` API more explicit
  Add `with_stdin` and `with_prefixed_stdin`, the former set `stdin` as is, the latter adds a LE u32 length prefix to the `stdin`, this is made for the `Platform::read_whole_input`
- `Platform::read_whole_input` now requires `stdin` to have a LE u32 length prefix to be more consistent, but will only returns the actual data part of it (as byte slice)
- Remove the `OutputHasher` generic of `Platform` implementation, and add a `OutputHashedPlatform` wrapper to replace it.
- Refactor docs
  - Update `README.md` ([rendered](https://github.com/eth-act/ere/blob/han/refactor/docs/README.md)), resolves https://github.com/eth-act/ere/issues/234
  - Add crate docs to all `ere-{zkvm}` to specify the dependencies
- Others
  - Remove the `protobuf-compiler` prerequisite, commit the generated `api.rs`, and use test to make sure the `api.rs` is synced with the proto file
  - Update `rust-version` to `1.88` in `Cargo.toml`

---

OpenVM is failing due to `svm-rs` error, need https://github.com/openvm-org/openvm/pull/2299 to be released to fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,18 +151,18 @@ Public values written in the guest program (via `Platform::write_whole_output()`
 
 Different zkVMs handles public values in different approaches:
 
-| zkVM      | Size Limit                | Note                          |
-| --------- | ------------------------- | ----------------------------- |
-| Airbender | 32 bytes                  | Padded to 32 bytes with zeros |
-| Jolt      | 4096 bytes (Configurable) |                               |
-| Miden     | 16 words                  |                               |
-| Nexus     | unlimited                 | Size configured automatically |
-| OpenVM    | 32 bytes                  | Padded to 32 bytes with zeros |
-| Pico      | unlimited                 | Hashed internally             |
-| Risc0     | unlimited                 | Hashed internally             |
-| SP1       | unlimited                 | Hashed internally             |
-| Ziren     | unlimited                 | Hashed internally             |
-| Zisk      | 256 bytes                 |                               |
+| zkVM      | Size Limit                | Note                            |
+| --------- | ------------------------- | ------------------------------- |
+| Airbender | 32 bytes                  | Padded to 32 bytes with zeros   |
+| Jolt      | 4096 bytes (Configurable) |                                 |
+| Miden     | 16 words                  | Word = Goldilocks field element |
+| Nexus     | unlimited                 | Size configured automatically   |
+| OpenVM    | 32 bytes                  | Padded to 32 bytes with zeros   |
+| Pico      | unlimited                 | Hashed internally               |
+| Risc0     | unlimited                 | Hashed internally               |
+| SP1       | unlimited                 | Hashed internally               |
+| Ziren     | unlimited                 | Hashed internally               |
+| Zisk      | 256 bytes                 |                                 |
 
 For zkVMs with size limits on public values, `OutputHashedPlatform<P, D>` serves as a wrapper that hashes outputs before calling the inner `P::write_whole_output`. This enables the same guest program to run across all zkVMs regardless of their size constraints:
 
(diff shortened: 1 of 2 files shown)


crates/zkvm-interface/src/zkvm.rs
```diff
@@ -49,6 +49,23 @@ impl Input {
...
+    ///
+    /// The `Platform::read_whole_input` requires stdin to have a LE u32 length
+    /// prefix for efficiency reason.
+    pub fn with_prefixed_stdin(mut self, stdin: Vec<u8>) -> Self {
+        self.stdin = Vec::with_capacity(4 + stdin.len());
+        self.stdin.extend((stdin.len() as u32).to_le_bytes());
+        self.stdin.extend(stdin);
+        self
+    }
+
```
han0110: If guest program uses `Platform::read_whole_input`, the host should use this method.

crates/dockerized/server/src/test.rs
```diff
@@ -0,0 +1,28 @@
...
+        .unwrap();
+
+    let latest = tempdir.path().join("api.rs");
+    let current = dir.join("src").join("api.rs");
+
+    // If it's in CI env, don't overwrite but only check if it's up-to-date.
+    if env::var_os("GITHUB_ACTIONS").is_none() {
+        fs::copy(&latest, &current).unwrap();
+    }
+    assert_eq!(fs::read(&latest).unwrap(), fs::read(&current).unwrap());
```
jsign: :+1:

crates/zkvm-interface/platform/src/lib.rs
```diff
@@ -54,3 +59,73 @@ pub trait Platform {
...
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        &self.0[4..]
+    }
+}
+
+impl LengthPrefixedStdin {
+    pub fn new(stdin: Vec<u8>) -> Self {
+        let len = u32::from_le_bytes(stdin[..4].try_into().unwrap());
```
jsign: Maybe paranoid comment but if `stdin` is shorter than 4 bytes, we'll panic. I think since this is internal it _shouldn't_ happen. Not sure if might be annoying to retun a `Result` -- no strong opinions.

Edit: saw `with_prefixed_stdin` below -- I guess if this is the only use should be okay.

jsign (approved):
LGTM!

jsign:
> OpenVM is failing due to svm-rs error, need https://github.com/openvm-org/openvm/pull/2299 to be released to fix.

Is this related to the 1.88 compilator bump in our repo?

crates/zkvm-interface/platform/src/lib.rs
han0110: > Uhm, this can also panic if whateverzkvmsdk::read_input() is shorter than 4 bytes -- but on properly used API this shouldn't happen. Maybe just worth an assert with a nicier panic msg, just to help someone misusing the api.

Yeah this makes sense, added some more panic message in https://github.com/eth-act/ere/pull/235/commits/a9f048966b6592f54c1a20281e0c8888a86900df.

README.md
```diff
@@ -13,154 +13,368 @@
...
 
-- `protobuf-compiler`
+Different zkVMs handles public values in different approaches:
 
-## Quick Start
+| zkVM      | Size Limit                | Note                          |
+| --------- | ------------------------- | ----------------------------- |
+| Airbender | 32 bytes                  | Padded to 32 bytes with zeros |
+| Jolt      | 4096 bytes (Configurable) |                               |
+| Miden     | 16 words                  |                               |
```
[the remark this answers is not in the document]
han0110: Added a note in https://github.com/eth-act/ere/pull/235/commits/a9f048966b6592f54c1a20281e0c8888a86900df.

PR #244, opened 2025-12-11
Refactors

han0110:
- Refactor `ere-dockerized` (mostly for `zkboost`)
  - Add crate `ere-common` and move `CompilerKind` and `zkVMKind` there
  - Rebuild base images only if the compiler and server is missing (the workload repo can pull only the compiler and server after this PR)
  - Add `ERE_IMAGE_REGISTRY` env variable detection and prepend to image if it's set (the workload repo doesn't need to re-tag but just setting `ERE_IMAGE_REGISTRY=ghcr.io/eth-act/ere`)
  - Derive serde for enums
- Refactor CI a bit
- Refactor dockerfiles a bit to reduce image size


jsign (approved):
Nice QOL improvements for us :+1: 
LGTM!

PR #262, opened 2026-01-03
dockerized: support ERE_DOCKER_NETWORK

jsign:
This PR adds support to detect the presence of `ERE_DOCKER_NETWORK`. If this env var is detected, it is assumed the current working environment is a Docker setup.

The value of the variable is the name of the docker network being used, which allows to assign to any spawned  `ere-server-*` container to join the same network, and connect properly by container name instead of `127.0.0.1` (which would be incorrect without forcing the context container to run in host network mode).

The motivation for this PR is using zkboost within a docker compose setup.

See https://github.com/eth-act/zkboost/pull/16 to get a bit more context!


han0110 (approved):
LGTM!

PR #263, opened 2026-01-04
Manage `sp1-gpu` container by `ere-sp1`

han0110:
After #262 we can set the docker network by env, but we couldn't set the network for the `sp1-gpu` container because it's managed by `sp1-sdk` (and there is no way to configure that)

This PR move the management of `sp1-gpu` container from `sp1-sdk` to `ere-sp1`, so we can set the docker network for it.

### DIFF
--- a/crates/dockerized/src/zkvm.rs
+++ b/crates/dockerized/src/zkvm.rs
@@ -1,11 +1,15 @@
 use crate::{
     compiler::SerializedProgram,
-    image::{base_image, base_zkvm_image, image_registry, server_zkvm_image},
+    image::{base_image, base_zkvm_image, server_zkvm_image},
     util::{
         cuda::cuda_arch,
         docker::{
             DockerBuildCmd, DockerRunCmd, docker_container_exists, docker_image_exists,
-            docker_pull_image, force_rebuild, stop_docker_container,
+            docker_pull_image, stop_docker_container,
+        },
+        env::{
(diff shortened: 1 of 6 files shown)


crates/dockerized/src/zkvm.rs
```diff
@@ -149,7 +153,7 @@ impl ServerContainer {
     ) -> Result<Self, Error> {
         let port = Self::PORT_OFFSET + zkvm_kind as u16;
 
-        let name = format!("ere-server-{zkvm_kind}-{port}");
```
han0110: Simplify the container name a bit.

crates/zkvm/sp1/src/zkvm/sdk.rs
```diff
@@ -1,9 +1,22 @@
...
+use std::{
+    env,
+    ops::Deref,
+    panic::{self, AssertUnwindSafe},
+    process::Command,
 };
+use tracing::error;
+
+// https://github.com/succinctlabs/sp1/blob/v5.2.4/crates/cuda/src/lib.rs#L207C34-L207C78.
+const SP1_CUDA_IMAGE: &str = "public.ecr.aws/succinct-labs/sp1-gpu:8fd1ef7";
```
jsign: Something else to remember (potentially) changing whenever SP1 SDK is updated. 😅

jsign (approved):
LGTM

PR #270, opened 2026-01-15
Support `--program-path` flag in `ere-server`

han0110:
- Support `--program-path` flag in `ere-server` so docker deployment of `ere-server` can be simplified a bit
- Simplify `zkVMClient` a bit
- Implement serde manually for `ProverResourceType`, so it can be field of untagged enum (for `serde_yaml` to work)
- Add `DockerizedzkVM::{execute,prove,verify}_async` for `zkboost` to integrate more easily

### DIFF
--- a/crates/zkvm-interface/src/zkvm/resource.rs
+++ b/crates/zkvm-interface/src/zkvm/resource.rs
@@ -3,7 +3,6 @@ use serde_untagged::UntaggedEnumVisitor;
 
 /// Configuration for network-based proving
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(rename_all = "kebab-case")]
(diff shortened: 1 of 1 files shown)


crates/dockerized/server/src/client.rs
```diff
@@ -6,61 +6,50 @@ use crate::api::{
...
-            }
-
-            match http_client.get(url.join("health").unwrap()).send().await {
-                Ok(response) if response.status().is_success() => break,
-                _ => sleep(INTERVAL).await,
-            }
-        }
-
-        let client = Client::new(url.join("twirp").unwrap(), http_client, Vec::new(), None);
+    pub fn new(endpoint: Url, http_client: reqwest::Client) -> Result<Self, Error> {
```
jsign: Which was the reason why we had this TIMEOUT, wasn't because server can take a bit to start?

PR #278, opened 2026-01-28
Updated ere-compiler to output ELF bytes and program digest separately

### LINKED ISSUE #275: Update `ere-compiler` to output ELF bytes and program digest separately
Currently the compiled artifact generated by `ere-compiler` is **not** just ELF bytes, but a bincode serialized structured defined by each zkVM, this makes it difficult for other projects to integrate when they only need to ELF part (e.g. https://github.com/lambdaclass/ethrex/pull/6019 creates a python3 script to extract the ELF part).

We can improve this by updating the `ere-compiler` cli to something like:

```bash
ere-compiler \
    --compiler-kind  rust-customized \
    --guest-dir      /path/to/guest \
    --output-dir     /path/to/output \
    --elf-name       guest.elf \       # output ELF to /path/to/output/guest.elf
    --digest-name    digest            # output serialized program digest to /path/to/output/digest when specified
```

The digest is the only essential part for verifier to verify a proof.

developeruche:
Closes #275

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3858,7 +3858,6 @@ dependencies = [
  "ere-zisk",
  "ere-zkvm-interface",
  "serde",
- "tempfile",
  "tracing",
  "tracing-subscriber 0.3.20",
 ]


crates/dockerized/compiler/Cargo.toml
```diff
@@ -28,6 +28,7 @@ ere-zkvm-interface.workspace = true
 ere-common.workspace = true
 
 [dev-dependencies]
+tempfile.workspace = true
```
han0110: seems not used (was it intended to be used in a test)?

Cargo.lock
```diff
@@ -3858,6 +3858,7 @@ dependencies = [
  "ere-zisk",
  "ere-zkvm-interface",
  "serde",
+ "tempfile",
```
han0110:
```suggestion
```

PR #286, opened 2026-02-04
feat(jolt): add secp256k1 inlines support 

brech1:
# Description

Adds `jolt-inlines-secp256k1` dependency to enable ECDSA in Jolt guests.

### DIFF
--- a/crates/zkvm/jolt/src/lib.rs
+++ b/crates/zkvm/jolt/src/lib.rs
@@ -47,5 +47,5 @@ pub mod zkvm;
 #[cfg(feature = "zkvm")]
 pub use zkvm::*;
 
-#[cfg(feature = "secp256k1-inlines")]
+#[cfg(feature = "zkvm")]
 use jolt_inlines_secp256k1 as _;


crates/zkvm/jolt/Cargo.toml
```diff
@@ -28,9 +29,10 @@ ere-test-utils = { workspace = true, features = ["host"] }
 ere-build-utils.workspace = true
 
 [features]
-default = ["compiler", "zkvm"]
+default = ["compiler", "zkvm", "secp256k1-inlines"]
 compiler = ["dep:ere-compile-utils"]
 zkvm = ["dep:jolt-sdk"]
+secp256k1-inlines = ["dep:jolt-inlines-secp256k1"]
```
han0110: Since the overhead is negligible, I think we could just always enable it

```suggestion
default = ["compiler", "zkvm"]
compiler = ["dep:ere-compile-utils"]
zkvm = ["dep:jolt-sdk", "dep:jolt-inlines-secp256k1"]
```

crates/zkvm/jolt/src/lib.rs
```diff
@@ -46,3 +46,6 @@ pub mod zkvm;
 
 #[cfg(feature = "zkvm")]
 pub use zkvm::*;
+
+#[cfg(feature = "secp256k1-inlines")]
```
han0110:
```suggestion
#[cfg(feature = "zkvm")]
```

PR #287, opened 2026-02-06
Add zisk cycle scope tracker

jsign:
This PR adds to `ZiskPlatform` proper "cycle tracking".

The reality is that this isn't about cycles but what Zisk calls "cost" which is related to AIR table surface. To know more about this see PR description https://github.com/0xPolygonHermez/zisk/pull/680.

This allows us to use `ziskemu` to see the actual cost per scope, which is more useful than cycles.

The main "tension" in this PR is that, because the Zisk API requires a `const` for the tag id, our `Platform` trait uses a `&str`.

The most obvious thing a priori was to hash it, but given that the id can only have up to 256 values I felt there was a high chance of collision.

I ended up taking a bit of a safer route, prioritizing avoiding collisions -- but I still don't like the linear search or the "name hashing" to a u64. Ideally, we would like these markers to have low cost in the guest program, and this feels like a stretch.

All this is to try to avoid changing the `Platform` trait -- if we change it, we could add a generic int constant to `cycle_scope_{start|end}` that gets used for Zisk and ignored in other zkVMs, but honestly, I find that somewhat ugly for UX since now the user has to ensure the IDs (which can be quite nested in real code between crates) do not overlap.

@han0110, leaving this as a draft to get a bit of your opinion :)

### DIFF
--- a/crates/zkvm/zisk/platform/src/lib.rs
+++ b/crates/zkvm/zisk/platform/src/lib.rs
@@ -23,9 +23,8 @@ fn hash_name(name: &str) -> u64 {
 /// Each unique scope name gets a unique tag ID (0, 1, 2, ...) on first use.
 /// Panics if more than 256 distinct scope names are registered.
 struct ScopeRegistry {
-    entries: UnsafeCell<[(u64, u8); 256]>, // (name_hash, tag_id)
+    entries: UnsafeCell<[u64; 256]>, // name hashes; tag = index
     count: UnsafeCell<u8>,
-    next_tag: UnsafeCell<u8>,
 }
 
 // SAFETY: ZiskPlatform runs in a single-threaded zkVM environment.
@@ -34,9 +33,8 @@ unsafe impl Sync for ScopeRegistry {}
 impl ScopeRegistry {
     const fn new() -> Self {
         Self {
-            entries: UnsafeCell::new([(0, 0); 256]),
+            entries: UnsafeCell::new([0; 256]),
             count: UnsafeCell::new(0),
-            next_tag: UnsafeCell::new(0),
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/zkvm/zisk/platform/src/lib.rs
```diff
@@ -2,13 +2,127 @@
...
+    }
+}
+
+static SCOPE_REGISTRY: ScopeRegistry = ScopeRegistry::new();
+
+/// Dispatches a runtime `u8` tag to a const-generic ziskos profiling function.
+/// Generates a 256-arm match to bridge the runtime value to compile-time const generics.
+macro_rules! dispatch_profile {
+    (start, $tag:expr) => {
+        dispatch_profile!(@start $tag,
```
jsign: Since Zisk ask for a const, we have a long match arm.

crates/zkvm/zisk/platform/src/lib.rs
```diff
@@ -2,13 +2,127 @@
...
+
+/// Global registry mapping scope name hashes to sequential tag IDs.
+///
+/// Each unique scope name gets a unique tag ID (0, 1, 2, ...) on first use.
+/// Panics if more than 256 distinct scope names are registered.
+struct ScopeRegistry {
+    entries: UnsafeCell<[(u64, u8); 256]>, // (name_hash, tag_id)
+    count: UnsafeCell<u8>,
+    next_tag: UnsafeCell<u8>,
+}
```
han0110: Perhaps we can keep one of `count` or `next_tag` since they are always equal? Also it seems we don't need to store the tag but just the `name_hash`, so it becomes a linear map with no allocation.

```suggestion
struct ScopeRegistry {
    entries: UnsafeCell<[u64; 256]>, // name_hash
    count: UnsafeCell<u8>,
}
```

crates/zkvm/zisk/platform/src/lib.rs
han0110: Was digging whether there is way to avoid this, but realized that [zisk uses `addi` as the mark of profiling](https://github.com/0xPolygonHermez/zisk/blob/main/core/src/riscv2zisk_context.rs#L114), so this makes sense!

han0110 (approved):
LGTM

Was thinking using `hashbrown::HashSet<u64, IdentityHasher>` with fnv pre-hash, but after doing some real benchmark (reserving the tag `0` for the `get_or_assign_tag`), I found that when key counts less than 35, using linear map is actually cheaper (both cycle count and memory usage), so perhaps current approach is good enough.

crates/zkvm/zisk/platform/src/lib.rs
jsign: Makes sense! Changed in https://github.com/eth-act/ere/pull/287/commits/12986a9984d8d2b0befb20b0146249fa059caab4

PR #289, opened 2026-02-07
Add macro `export_cycle_scope_names` in `ere-platform-zisk`

han0110:
For `ziskemu` to display human-readable cycle scopes, we can do:

```rust
// main.rs

#![no_main]

use ere_platform_zisk::{export_cycle_scope_names, ziskos, ZiskPlatform};

ziskos::entrypoint!(main);

fn main() {
    Guest::run::<ZiskPlatform>();

    export_cycle_scope_names!(
        cycle_scope_a,
        cycle_scope_b,
        cycle_scope_c,
        cycle_scope_d,
        cycle_scope_e,
    );
}
```

To make sure it matches what's registered during `Guest::run`, we can enable the feature `check-cycle-scope` of `ere-platform-zisk` like:

```toml
# Cargo.toml

ere-platform-zisk = { workspace = true, features = ["check-cycle-scope"] }
```

If the names in `export_cycle_scope_names!(...)` mismatch the actual registered ones, panic message like the following will be displayed (for copy-paste to `main.rs`):

```
Cycle scope names mismatch with the registered entries, expect:

export_cycle_scope_names!(
    cycle_scope_f,
    cycle_scope_g,
    cycle_scope_h,
    cycle_scope_i,
    cycle_scope_j,
);
```

### DIFF
--- a/crates/zkvm/zisk/platform/src/lib.rs
+++ b/crates/zkvm/zisk/platform/src/lib.rs
@@ -199,11 +199,13 @@ pub fn check_cycle_scope_names(_names: &[&str]) {
 
 /// Exports cycle scope names as symbols for `ziskemu` to display human-readable names.
 ///
-/// Names must be listed in the same order as they were first used via
-/// [`Platform::cycle_scope_start`].
+/// The declared scope names must be declared in the order they are _executed_
+/// in your code. Not doing so can dangerously name a scope incorrectly. Be
+/// careful if you have lambda functions or conditionals since it is easy to
+/// confuse the right order.
 ///
 /// Enable feature `check-cycle-scope` to verify correctness at runtime
-/// (panics on mismatch with expected output to copy-paste).
+/// automatically (panics on mismatch with expected output to copy-paste).
 #[macro_export]
 macro_rules! export_cycle_scope_names {
     ($($name:ident),* $(,)?) => {{


crates/zkvm/zisk/platform/src/lib.rs
```diff
@@ -161,3 +184,73 @@ impl Platform for ZiskPlatform {
...
+             224 225 226 227 228 229 230 231 232 233 234 235 236 237 238 239
+             240 241 242 243 244 245 246 247 248 249 250 251 252 253 254 255]
+            [] [$($name)*]
+        );
+    }};
+}
+
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __export_cycle_scope_names {
```
jsign: Might be worth adding some comment to the macro to warn users something like:
```
/// The declared scope names must be declared in the order they are
/// _executed_ in your code. Not doing so can dangerously name a scope
/// incorrectly. Be careful if you have lambda functions or conditionals
/// since it is easy to confuse the right order.
```
I experienced this myself since the first time I used it, I named this in the way  I knew they appeared in the "main flow". But one of the function calls had a lambda-function which also declared some scope, which shifted all the results.

A priori by looking at the ziskemu output I saw nice scopes with their metrics, but they were actually shifted by 1 place. I realized this easily since clearly I was seeing a scope with a very high cost with a low one and viceversa.

I think leaving the comment is enough, since we are already jiggling balls to support this. If the user wants more control, they can always use `ziskos` directly. :)

PR #290, opened 2026-02-09
feat(jolt): add keccak and sha2 inlines

brech1:
# Description

Adds support for jolt `keccak` and `sha2` inlines.


han0110 (approved):
LGTM!

PR #291, opened 2026-02-10
Build and publish cuda enabled docker image

han0110:
- Rename env `CUDA_ARCH` to `CUDA_ARCHS` to accept comma-separated numeric part of CUDA compute capabilites e.g. `CUDA_ARCHS=89,120`, if the zkVM supports multiple CUDA compute capabilites it'll be forwarded (Airbender, OpenVM, Risc0), if not the largest will be used (ZisK)
- Update dockerfiles to use the env var that the zkVM recognizes, and translate `CUDA_ARCHS` to that when building image
- Update CI
  - For each PR, it additionally tests building CUDA enabled images using `CUDA_ARCHS=89,120` (L40S, RTX 40 series, RTX 50 series), this works fine for most zkVMs but takes a bit long for Risc0 (~1hr), but it'd be cached in future PR if the dockerfile is untouched.
  - For push to master it builds and pushs images with git sha tag (without and with CUDA enabled if supported), also builds and pushs cluster image if supported. After build and push succeeds, the test zkVM workflow will be triggered to run same test as PR but using the published images.
  - For push to tag, it add the semver tag on the git sha tag for all images pushed.

### DIFF
--- a/crates/dockerized/src/zkvm.rs
+++ b/crates/dockerized/src/zkvm.rs
@@ -35,7 +35,7 @@ use std::{
 };
 use tempfile::TempDir;
 use tokio::{sync::RwLock, time::sleep};
-use tracing::{error, info};
+use tracing::{error, info, warn};
 
 mod error;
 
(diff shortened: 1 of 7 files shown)


crates/dockerized/src/zkvm.rs
```diff
@@ -41,6 +41,40 @@ mod error;
 
 pub use error::Error;
 
+/// Applies per-zkVM CUDA architecture build args to a Docker build command.
```
jsign: Maybe obvious question to understand the intention of the PR: if a machine signals support for 89 and 120, why would they care about 89 and not always select the highest? (as Zisk does apparently)
I was thinking of multi-GPU machines with different kinds of GPUs, but that would mean Zisk would work there.

PR #107, opened 2025-08-26
Add rust stock compiler.

rodiazet:
This PR is a draft and WiP of stock rust compilation of a guest program for all integrated zkvms.

PR #296, opened 2026-02-19
Update ZisK to `0.16.0`

### LINKED ISSUE #73: Use ZisK's SDK
ZisK's SDK is WIP here https://github.com/0xPolygonHermez/zisk/pull/355.

Currently in `ere-zisk` we uses `cargo-zisk` to do everything, but it'd be easier to work with a SDK, and easier to track the version as well.

### LINKED ISSUE #308: Update `zisk` to `v0.16.0`
A new version of `zisk` is available.

- Current version: `v0.15.0`
- Latest version: `v0.16.0`

han0110:
- Update zisk to `v0.16.0`. Resolves #308
  - Patch is still needed, the diff can be found https://github.com/0xPolygonHermez/zisk/compare/v0.16.0...han0110:patch/v0.16.0, by commits:
    - https://github.com/han0110/zisk/commit/8e5b3eec1fa2cb89788a0988a61fc7ecc7db9c03 - Set complete for the ASM service so there is no dangling threads that causes memory leak, without this the prover restart will end up out of memory
    - https://github.com/han0110/zisk/commit/44cd694df931cc4e89898f5544f87245775d7e6f - Allow passing data from client, previously the data must be in path accessible by the coordinator
    - https://github.com/han0110/zisk/commit/21524683f0f6c3e111dc04f42e8fde805f44b2f0 - Add gRPC endpoint `SubscribeToProof` to allow subscribe to proof completion/failure update, previously it only supports webhook
    - https://github.com/han0110/zisk/commit/61fe43922231290ec0e09f482a88bbf3572a8f1c - Disable local request check, previously it only allows request from localhost
- Use `zisk-sdk` instead of using `cargo-zisk`. Resolves #73
  - No need to setup explicitly, the prover now runs setup automatically if it's setup yet
  - Use local prover instead of server, recovery from failure by cleaning up ASM services manually
- Add stock rust `Compiler` impl `RustRv64ima` with target spec as json (supercedes #107, which has ZisK's stock rust compiler left unmerged)

### DIFF
--- a/crates/zkvm/zisk/src/zkvm/sdk/local.rs
+++ b/crates/zkvm/zisk/src/zkvm/sdk/local.rs
@@ -6,7 +6,7 @@ use blake3::Hash;
 use ere_zkvm_interface::zkvm::CommonError;
 use parking_lot::{Mutex, MutexGuard};
 use proofman_common::ParamsGPU;
-use std::{env, panic, path::PathBuf, process::Command, thread::sleep, time::Duration};
+use std::{env, fs, panic, path::PathBuf, process::Command, thread::sleep, time::Duration};
 use tempfile::tempdir;
 use tracing::info;
 use zisk_rom_setup::generate_assembly;
@@ -130,6 +130,9 @@ fn initialize(
     elf_hash: Hash,
 ) -> Result<(ZiskProver<Asm>, ZiskProgramPK), Error> {
     info!("Initializing ZisK prover...");
+
+    fs::create_dir_all(cache_dir())
+        .map_err(|err| CommonError::create_dir("cache", cache_dir(), err))?;
 
     if !assembly_files_exist(elf_hash) {
         generate_assembly(elf, ELF_NAME, &cache_dir(), false, false)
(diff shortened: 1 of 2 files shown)


crates/zkvm/zisk/src/zkvm/sdk.rs
```diff
@@ -84,265 +79,210 @@ impl ZiskSdk {
...
+        let cap = aligned.capacity() * size_of::<u64>();
+        // SAFETY: `ptr` came from a `Vec<u64>` allocation.
+        unsafe { Vec::from_raw_parts(ptr, len, cap) }
     })
 }
 
+/// Returns `data` with a LE u64 length prefix and padding to multiple of 8.
+///
+/// The length prefix and padding is expected by ZisK emulator/prover runtime.
+fn length_prefixed_and_padded(data: &[u8]) -> Vec<u8> {
```
han0110: Now ZisK requires input to be prefixed with u64 LE length, and padded to multiple of 8

PR #298, opened 2026-02-23
Update `miden` to `v0.21.0`

### LINKED ISSUE #282: Update `miden` to `v0.21.0`
A new version of `miden` is available.

- Current version: `v0.20.4`
- Latest version: `v0.21.0`

han0110:
In this version of miden, it replaces Winterfell with Plonky3, hence the field API changes a bit. Resolves #282

### DIFF
--- a/crates/zkvm/miden/src/zkvm.rs
+++ b/crates/zkvm/miden/src/zkvm.rs
@@ -151,6 +151,7 @@ impl zkVM for EreMiden {
             Deserializable::read_from_bytes(proof)
                 .map_err(|err| CommonError::deserialize("proof", "miden", err))?;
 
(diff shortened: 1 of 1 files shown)


crates/zkvm/miden/src/zkvm.rs
```diff
@@ -149,15 +151,12 @@ impl zkVM for EreMiden {
             Deserializable::read_from_bytes(proof)
                 .map_err(|err| CommonError::deserialize("proof", "miden", err))?;
 
+        // Security level is hardcoded as 96, so we skip the check.
+
```
brech1: Perhaps we could add where this happens if we're not asserting? 

- https://github.com/0xMiden/miden-vm/blob/6623397867b6bd0cf316f07e0429fb377345537a/air/src/config/blake3.rs#L81

PR #304, opened 2026-03-04
Update Jolt to revision 35d46f5 (with ZK support)

alxkzmn:
- Updated jolt dependencies in Cargo.toml to use the latest commit (35d46f5).
- Added new feature `zk` in `crates/zkvm/jolt/Cargo.toml` for better modularity.
- Modified platform features to include `jolt-common/std`.
- Refactored memory configuration in `lib.rs` to replace MEMORY_SIZE with HEAP_SIZE.
- Updated compiler to use DEFAULT_HEAP_SIZE instead of DEFAULT_MEMORY_SIZE.
- Adjusted linker script to define heap boundaries.
- Changed error handling in `error.rs`.
- Updated SDK to use heap size instead of memory size and adjusted related configurations.
- Set default Rust toolchain to 1.88 in Dockerfile for consistent builds.

### DIFF
--- a/docker/jolt/Dockerfile.compiler
+++ b/docker/jolt/Dockerfile.compiler
@@ -19,6 +19,8 @@ FROM $RUNTIME_IMAGE AS runtime_stage
 RUN apt-get update && apt-get install -y --no-install-recommends \
     build-essential \
     ca-certificates \
+    cmake \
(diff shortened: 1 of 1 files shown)


han0110 (approved):
LGTM

PR #350, opened 2026-05-03
ci(risc0): pass GITHUB_TOKEN as secret env to fix rzup rate limits

### LINKED ISSUE #41: CI: jobs usually hit rate limiting
Our current CI pipeline seems quite heavy for unauthenticated GH API calls, so jobs sometimes fail with:
```
2.598 error: Remote error: api.github.com: API rate limit exceeded for 172.191.240.39. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```

This can be worked around by retrying manually.

See [this run](https://github.com/eth-act/ere/actions/runs/15823648787/job/44598413877?pr=40#step:5:86) as an example.

If we authenticate calls, we can have a higher rate limit while still in the free plan.

Andrurachi:
Fixes #41

Injects GITHUB_TOKEN into the Risc0 Docker build via BuildKit secret mount (env=GITHUB_TOKEN). This prevents the rzup installer from hitting GitHub API rate limits in CI.

### DIFF
--- a/.github/scripts/build-image.sh
+++ b/.github/scripts/build-image.sh
@@ -179,59 +179,53 @@ if [ -n "$RUSTFLAGS" ]; then
     SERVER_ZKVM_BUILD_ARGS+=(--build-arg "RUSTFLAGS=$RUSTFLAGS")
 fi
 
-# Pass GITHUB_TOKEN as a BuildKit secret if available to prevent rzup rate limits
-SECRET_ARGS=()
-if [ -n "${GITHUB_TOKEN:-}" ]; then
-    SECRET_ARGS=(--secret id=github_token,env=GITHUB_TOKEN)
+# Pass GITHUB_TOKEN to prevent rzup hit github rate limits
+if [ -n "$GITHUB_TOKEN" ] && [ "$ZKVM" = "risc0" ]; then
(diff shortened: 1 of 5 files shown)


Andrurachi:
Hey @han0110 ,

It looks like the token fix successfully bypassed the rate limit, but it caused ziskup to fetch the latest unpinned version of ZisK (v0.16.1).

It looks like v0.16.1 broke the Docker builds for two reasons:
- ziskup now triggers an interactive prompt asking to select CPU vs GPU, which crashes the CI runner (/dev/tty: No such device or address).
- The binary structure seems to have changed, breaking the mv cargo-zisk-gpu command in the install script.

I see you just opened #351 to update ZisK to v0.17.0 (which I assume handles these changes). Once your PR is merged, I can rebase this branch on top of it. Let me know if you'd prefer me to handle this differently in the meantime :)

han0110 (commented):
Checked the last 3 months workflow github rate limiting failures, all come from `rzup` installation, which only requires pass `GITHUB_TOKEN` to risc0's installation script:

```
RUN --mount=type=secret,id=github_token,env=GITHUB_TOKEN \
    /tmp/install_risc0_sdk.sh && rm /tmp/install_risc0_sdk.sh
```

The rest changes seem not necessary and overkill given the complexity.

han0110 (approved):
Thanks for the PR!

PR #351, opened 2026-05-04
feat: update `zisk` to `v0.17.1`

han0110:
Pending for `v0.17.1` release which fixes the GPU proving on cuda 12.9.

---

- Update ZisK to `v0.17.0`, a patch is still needed to recover from prover error
  - https://github.com/han0110/zisk/commit/9499da6cad4413f264a3170f57fd80cae4bd0e40
  Set complete for the ASM service so there is no dangling threads that causes memory leak, without this the prover restart will end up out of memory
  - https://github.com/han0110/zisk/commit/4c380594c92680f0e76329f19f8c3e13045dd669
  Expose a function to cleanup ASM services when it fails and exits (e.g. guest program panics)
  - https://github.com/han0110/zisk/commit/6d09f8e4b6acbf04a1413654f6cd0b2c1887872b
  Expose a function to compute program vk without starting the ASM services
  - https://github.com/han0110/zisk/commit/4019fa3ef32cabef4c93188029956c1efaba6654
  Patch the `ziskfloat.elf` by extracting from the released `cargo-zisk` in upstream repo, and disable the `build.rs` script of `lib-float` crate to avoid `ziskfloat.elf` gets updated (which makes the program vk different). This ensures the ZisK prover in Ere to compute the same program vk as the official released `cargo-zisk` does.
- Update the `ZiskProof` format to match the format expected by `proofman-verifier`, and do strict constant size check to avoid the verifier panics
- Update the `ZiskPlatform` implementation, now the profile syscall reads string as naming instead of requiring a constant tag. However, the [AOT compilation process doesn't support ELF with the profile syscall](https://github.com/0xPolygonHermez/zisk/blob/v0.17.0/core/src/zisk_rom_2_asm.rs#L7593-L7595) (depended by prover), so currently the `cycle_scope_*` functions are left no-op.
  
  @jsign This breaks the zisk profilng flow in `zkevm-benchmark-workload`, we probably need a custom build for the guests to do profiling, or we adapt to the [Function-Level Profiling](https://github.com/0xPolygonHermez/zisk/blob/v0.17.0/book/developer/profiling.md#function-level-profiling) that works for prover and enabled by default like:

  <details>
  <summary><code>ziskemu -e &lt;elf-path&gt; -i &lt;input-path&gt; -X -S -D</code></summary>

  ```
  ...

    TOP STEP FUNCTIONS (STEPS, % STEPS, CALLS, STEPS/CALL, FUNCTION)
  ----------------------------------------------------------------
              475   3.93%          1             475 <ere_platform_zisk::platform::ZiskPlatform as ere_platform_core::platform::Platform>::write_whole_output
              474   3.92%          1             474 <ere_util_test::program::basic::BasicProgramOutput<ere_util_test::codec::BincodeLegacy> as ere_codec::encode::Encode>::encode_to_vec
              470   3.89%          1             470 ziskos::io::commit_slice
              443   3.67%          1             443 <ere_util_test::program::basic::BasicProgramOutput<…> as serde_core::ser::Serialize>::serialize::<bincode::features::serde::ser::SerdeEncoder<…>>
              291   2.41%          3              97 <alloc::raw_vec::RawVecInner<_>>::reserve::do_reserve_and_handle::<alloc::alloc::Global>
              241   1.99%          1             241 <ere_util_test::program::basic::BasicProgramInput<ere_util_test::codec::BincodeLegacy> as ere_codec::decode::Decode>::decode_from_slice
              206   1.70%          1             206 <bincode::features::serde::de_borrowed::SerdeDecoder<…> as serde_core::de::Deserializer>::deserialize_struct::<<…>::deserialize::__Visitor<…>>
              184   1.52%          3              61 <alloc::raw_vec::RawVecInner>::finish_grow
               63   0.52%          1              63 <ere_platform_zisk::platform::ZiskPlatform as ere_platform_core::platform::Platform>::read_whole_input
               42   0.35%         14               3 memcpy
               40   0.33%          2              20 __rustc::__rust_realloc
               40   0.33%          1              40 ziskos::io::read_input_slice
               32   0.26%          2              16 __rustc::__rust_alloc
               16   0.13%          1              16 __rustc::__rust_alloc_zeroed
                3   0.02%          3               1 __rustc::__rust_dealloc
                3   0.02%          3               1 __rustc::__rust_no_alloc_shim_is_unstable_v2
                2   0.02%          2               1 zkvm_init

  ...
  ```

  </details>
- Update the `ZiskClusterClient` API, adds function `create_prove_job`, `wait_prove_job`, and `cancel_prove_job`, and `prove` is built by composing the async endpoints with an optional deadline.


crates/prover/zisk/Cargo.toml
```diff
@@ -34,10 +29,11 @@ ere-verifier-zisk.workspace = true
 [dev-dependencies]
 ere-compiler-zisk.workspace = true
 ere-util-test = { workspace = true, features = ["host"] }
+tempfile.workspace = true
 
 [features]
 default = []
-cuda = ["zisk-sdk/gpu"]
+cuda = []
```
han0110, line 36: In `v0.17.0` ZisK has cuda enabled by default (by detecting if `nvcc` is available), and a `cpu-only` feature flag is available to force cuda disabled. If we try to initialize gpu prover without `nvcc` available, it'd show a runtime error, so we add a `build.rs` to turn that into a compile time error.

crates/verifier/zisk/src/proof.rs
```diff
@@ -1,57 +1,107 @@
...
+            )));
+        }
+        if self.0[0] != PROOF_PREFIX_WORDS as u64 {
+            return Err(Error::InvalidProofFormat(format!(
+                "proof n_publics is {}, expected to be {PROOF_PREFIX_WORDS}",
+                self.0[0],
+            )));
+        }
+        Ok(())
+    }
```
han0110, line 104: Always validates the constant proof size and metadata, otherwise the verify function in `proofman-verifier` might be tricked to panic.

docker/zisk/Dockerfile.compiler
```diff
@@ -21,6 +21,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     ca-certificates \
     nasm \
     libgmp-dev \
+    libclang-dev \
+    clang \
```
han0110, line 25: Added for ZisK's `zkvm-interface` crate compilation.

docker/zisk/Dockerfile.server
```diff
@@ -53,7 +53,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     libclang-dev \
     clang \
     # Used to kill the ASM services when prover panics
-    psmisc \
+    procps \
```
han0110, line 56: We use `pkill` for defensive ASM service cleanup now (previously using `fuser`)

crates/prover/zisk/src/sdk/local.rs
```diff
@@ -43,198 +47,137 @@ impl Config {
...
-        let (proof, proving_time) = match result {
-            Err(err) => {
-                uninitialize(guard);
-                Err(Error::ProvePanic(panic_msg(err)))
-            }
+        let started = Instant::now();
+        let result = panic::catch_unwind(AssertUnwindSafe(|| {
+            self.prover
+                .prove(&self.program, stdin)
+                .wrap_proof(ProofKind::VadcopFinalMinimal)
```
han0110, line 113: Wrap into the minimal stark proof for smaller proof size (~256 KiB instead of ~329 KiB), this is also the default proof kind.

crates/cluster-client/zisk/src/client.rs
```diff
@@ -182,59 +214,43 @@ fn framed_stdin(data: &[u8]) -> Vec<u8> {
...
-        .copied()
-        .ok_or_else(|| Error::InvalidProofFormat("proof values are empty".to_string()))?
-        as usize;
-    if n_publics != N_PUBLICS {
-        return Err(Error::InvalidProofFormat(format!(
-            "unexpected publics word count: expected {N_PUBLICS}, got {n_publics}"
-        )));
+fn parse_proof(bytes: &[u8]) -> Result<ZiskProof, Error> {
+    #[derive(Default, Serialize, Deserialize)]
+    pub struct Proof<'a> {
```
jsign, line 219: Are all these structs not exported by Zisk?

crates/platform/zisk/src/platform.rs
```diff
@@ -34,13 +32,23 @@ impl Platform for ZiskPlatform {
         }
     }
 
-    fn cycle_scope_start(name: &str) {
-        let tag = SCOPE_REGISTRY.get_or_assign_tag(name);
-        dispatch_profile!(start, tag);
+    fn cycle_scope_start(_name: &str) {
```
jsign, line 37: > @jsign This breaks the zisk profilng flow in zkevm-benchmark-workload, we probably need a custom build for the guests to do profiling, or we adapt to the [Function-Level Profiling](https://github.com/0xPolygonHermez/zisk/blob/v0.17.0/book/developer/profiling.md#function-level-profiling) that works for prover and enabled by default like:

Uhm, function level profiling is a bit annoying since doesn't allow to undesrtand more cleanly where the cost is comming from in the app level logic.

By "custom build" you mean compiling the guest program or running it with Zisk with some feature enabled, or you mean actually re-implementing all the logic to have custom scopes as they had before? 

Got a bit lost since you also mention they now support string scope names instead of constants, so "scope profiling" is still supported in some official way?

crates/prover/zisk/src/sdk/local.rs
```diff
@@ -43,198 +47,137 @@ impl Config {
...
+        .map_err(Error::BuildProver)
 }
 
-fn uninitialize(mut prover_and_pk: MutexGuard<Option<(ZiskProver<Asm>, ZiskProgramPK)>>) {
-    info!("Uninitializing ZisK prover...");
-
-    let _ = Command::new("fuser")
-        .args(["-k", "-9", "23115/tcp", "23116/tcp", "23117/tcp"])
+/// Clear the program cache so the next `setup` spawns fresh ASM services, then SIGTERM any orphan
+/// ASM child still bound to our `ZISK_{pid}_*` shmem prefix.
```
jsign, line 164: Tangent question: do you feel the cluster setup is more elegant in how setup/teardown works?

crates/verifier/zisk/src/proof.rs
jsign: Uhm, panicking on verifying untrusted bytes sounds more like a bug to me. Prob we should ask them to avoid this eventually.

jsign (approved):
LGTM, left just some comments/questions.

crates/cluster-client/zisk/src/client.rs
han0110: Yes, there is structure for the proof, but the proof is eventually flatten into the format added in this PR, and in the verifier they don't check the size, so a malformed proof will trigger `index out of bounds`.

Will file an issue in upstream as well.

crates/platform/zisk/src/platform.rs
han0110: > By "custom build" you mean compiling the guest program or running it with Zisk with some feature enabled, or you mean actually re-implementing all the logic to have custom scopes as they had before?

custom build I mean build with `ziskos::ziskos_syscall!(ziskos::SYSCALL_PROFILE_ID, ...)` uncommented, it's commented out becuase with these syscalls, the ELF can't be proved (but can be used with `ziskemu` to do the profiling we used to do)

> Got a bit lost since you also mention they now support string scope names instead of constants, so "scope profiling" is still supported in some official way?

Scope profiling is supported by `ziskemu`, but the ELF can't be proved anymore.

crates/prover/zisk/src/sdk/local.rs
han0110: I checked the teardown logic in zisk and it looks like there are still some chance the ASM services get dangling, and the start function doesn't cleanup them, so I would prefer to keep the `pkill` for now

crates/platform/zisk/src/platform.rs
jsign: Oh, I see. so the `ziskos` crate that teh guest program needs to add these scope markers needs a fork that uncomment some code. And we can use normal ere code (i.e non-modified) to emulate the elf run?

jsign: If that's the case, ere and workload isn't strictly affected but more like the guest program `ziskos` dependency used? 

This assuming that we put the markers manually in guest with `ziskos`.

If we rely on ere for cycle-scoping, I agree that prob means custom ere which maybe is the most elegant thing now that strings names are supported?

crates/prover/zisk/src/sdk/local.rs
jsign: Ah but I was wondering in cluster mode -- that setup also needs these kind of services teardown stuff? (mainly asking for curiosity)

crates/platform/zisk/src/platform.rs
han0110: > This assuming that we put the markers manually in guest with `ziskos`.

I thought this might be the solution, but I just had another idea: Add a feature flag in `ere-platform-zisk` to enabling these profiling, so we can keep the same guest code and no forking, just need to enable a feature flag to generate the ELF for profiling only.

Will add a commit soon

han0110: Added in https://github.com/eth-act/ere/pull/351/commits/1810ac4e0ccbc8690e742edbe966c48bd52d7c88, so I think in `ere-guests` we can release another ELF with the feature on for the profiling (need to change the ere-compiler to accept feature flags first, will add a follow up PR soon)

jsign: Sounds good :pray:

crates/prover/zisk/src/sdk/local.rs
han0110: Ah I see, just tested and it seems if the worker ASM services die (guest program panics), it'll remain in bad status and coordinator doesn't remove the worker from the pool, so the whole cluster is also paralyzed...

Will need to figure out a way to check if the worker is still healthy so the supervisor can restart it if needed (e.g. `docker-compose` health check by grepping the asm services).

han0110:
Superseded by #367

Issue #28, opened 2025-05-29
Separate Proving time + execution time


kevaundray:
When given the proving time, it would be great to also know how long execution took to run

jsign:
@kevaundray is it possible to also add the cycles when doing proving mode? IIRC, when I saw the output in the prover machine the cycles weren't reported apparently in prover mode.

kevaundray:
Ah missed this, some of the zkVMs don't return the number of cycles, though it should be available

Andrurachi:
Hi! I'm currently working on this one. Just wanted to make sure since this is an old issue it is still planned :)

han0110:
Assigned, thanks!

PR #358, opened 2026-05-09
feat(provers): Add optional cycle tracking to proving report

Andrurachi:
Resolves part of #28.

Updates `ProgramProvingReport` to track cycle counts during proving mode. `execution_duration` was omitted because no current zkVM SDK cleanly separates or exposes internal execution time during a `prove()` call (due to fused/black-box trace generation pipelines).

**Changes:**

* **`core`**: Added `total_num_cycles: Option<u64>` to `ProgramProvingReport`.
* **`risc0`**: Populated cycles via the exposed `prove_info.stats.total_cycles`.
* **`airbender`**: Extracts cycles directly from the GPU prover's `ProveResult`. (The interpreter's pre-flight pass is not timed to emulate execution_duration. Execution is fused into the proving pipeline, making pre-flight timing an inaccurate reflection of actual proving-phase execution.
* **`sp1` / `openvm` / `zisk`**: Default to `None`. Their respective SDKs drop or hide execution metrics during proving, returning only the final proof artifacts. (To get cycles for these, `execute()` must be call prior to `prove()`).

### DIFF
--- a/crates/prover/airbender/src/prover.rs
+++ b/crates/prover/airbender/src/prover.rs
@@ -149,13 +149,13 @@ impl zkVMProver for AirbenderProver {
             .map_err(|err| Error::ExecutePanic(panic_msg(err)))??;
 
         let start = Instant::now();
-        let (proof, receipt) = match gpu_prover.prove(&input_words)? {
+        let (proof, receipt, cycles) = match gpu_prover.prove(&input_words)? {
             ProveResult {
                 proof: Proof::Real(proof),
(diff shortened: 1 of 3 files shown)


han0110 (approved):
LGTM, tho I don't think this resolves the issue, it might still be a nice thing to have, thanks for the PR.

I think what we want is the execution time of the whole proving time, which none of the zkVM provides yet. (unassigned you from the issue)

PR #393, opened 2026-07-29
feat: add verification key generation document

han0110:
- Show OpenVM version as `2.0.0` in `ere-catalog` (it was using git sha due to fork)
- Update OpenVM to use `bitcode` which is used in OpenVM for other objects (tho the official `cargo openvm commit` outputs json, but we need custom script to do the vk generation anyway)
- Update SP1 program VK to encode to `bytes32` representation to match the CLI output
- Add `docs/vk-generation.md` for guest teams to take reference.


docs/vk-generation.md
```diff
@@ -0,0 +1,161 @@
...
+
+## OpenVM
+
+### Format
+
+`bitcode` encoding of the upstream `VerificationBaseline`.
+
+### Instructions
+
+`cargo openvm commit` has no flag for a compiled ELF and works off a Rust project it builds itself, so the key comes from the rust script below instead. The script applies the same app config `ere-prover-openvm` builds its prover with and generates the aggregation keys in process, so no OpenVM installation is involved.
```
han0110: Have to use Rust scripting because OpenVM assumes guest to be Rust project, can't simply assign ELF path to it to do VK generation.

docs/vk-generation.md
```diff
@@ -0,0 +1,161 @@
+# Verification Key Generation
+
+Ere checks a proof against a program verifying key that identifies one specific compiled guest ELF. `Verifier::new` in `ere-verifier` takes that key as an opaque byte string and decodes it with the codec of the selected zkVM, so a key file is only accepted when it already matches that exact wire encoding.
```
jsign: Nitpick: Maybe we can avoid saying "key" and "key file" and always say "VK" (or just always the same word)?

docs/vk-generation.md
jsign: Ah, I see. I think at some point we can ask OpenVM to accept `--elf` as part of `cargo openvm commit` to simplify this as much as possible.

When that exists, we can simplify this document accordingly.

jsign (approved):
LGTM!

This will be quite useful for teams. I'll include it in the guest handbook when merged!

docs/vk-generation.md
jsign, line 1: Ah noticed the naming of `.vk` wasn't clarified in guest book ([now added](https://github.com/eth-act/zkevm-standards/pull/38/changes/ef32c5087ef7a37df8f16fe5c3c6744912a4f164)) and in this doc we don't include the zkVM version which sounds like a good idea?

docs/vk-generation.md
han0110: Make sense, will ask that

docs/vk-generation.md
han0110: Addded in https://github.com/eth-act/ere/pull/393/commits/575349924245f930e8b4bdc246a0914ef86cd8b2

docs/vk-generation.md
han0110: Fixed in https://github.com/eth-act/ere/pull/393/commits/c3e283875d86a28a7362984c054ab76da11346f2

PR #424, opened 2026-09-15
Demo, do not merge: a declared length aborts the decoder

frisitano:
> **Demonstration only. Not for merge, and CI is expected to fail on this branch.** The fix is in #423, which carries the same test.

# Description

This branch adds a test and nothing else, so that the bug can be seen on its own before reading a fix.

`impl_codec_by_bincode_legacy!` decodes with `bincode::config::legacy()`, which carries no limit. Under that configuration a length prefix in the input decides how much a container allocates, and the allocation happens before any of that length has been read.

Twelve bytes are enough to reach it through `SP1Proof`:

```rust
let mut input = 3u32.to_le_bytes().to_vec();
input.extend_from_slice(&211_946_530_762_463_256u64.to_le_bytes());
SP1Proof::decode_from_slice(&input);
```

```console
$ cargo test -p ere-verifier-sp1 --test decode_limit
running 2 tests
memory allocation of 211946530762463256 bytes failed
test malformed_input_is_rejected ... ok
error: test failed ... (signal: 6, SIGABRT: process abort signal)
```

The assertion in the test is never reached. The length is representable, so it is not caught as a capacity overflow; it becomes a real allocation request, the allocator fails, and an allocation failure calls `abort()`. That is not a panic, so a caller cannot catch it either.

Both fields are chosen by whoever supplies the bytes, so this is reachable wherever proof bytes arrive from a network. The `UnexpectedProofKind` check that would reject this input runs on the decoded value, so it never gets to run.

## Why it is this input and not any input

The second test is included for contrast and passes on this branch. Arbitrary bytes, an empty input and a truncated proof are all rejected cleanly, because their leading word is not a selector this decoder accepts, so no length is ever read.

The selector used above is the one an OpenVM stateless-validator proof begins with, `03 00 00 00`, which `ProofFromNetwork` also accepts. I found this feeding a published OpenVM proof to the SP1 verifier; the other five cross-pairings between the three zkVMs are rejected cleanly. So it is specifically the case where one format's opening word is a selector another format admits.

`ZiskProof` uses the same macro and so shares the exposure, though I do not have an input that reaches it.

## Next

#423 bounds the decode, which turns this input into `DecodeError::LimitExceeded`, and carries this same test plus a check that the bound has not narrowed what is accepted.


han0110:
Closed by #423

